Issue #2, opened 2018-02-12
"started_tags" and "stopped_tags" not being set on scheduled actions


CliveJL:
I've just setup the AWS Instance Scheduler and configured a single instance with a "Schedule" Tag to test it out. The schedule seems to be working fine; the instance is being Stopped and Started according to the schedule defined in DynamoDB.

However, the "Automated Tagging" feature doesn't seem to be working. The DynamoDB config looks correct; I simply copied and pasted the tag format from the documentation, and it appears in the "config" table as below:

```
"started_tags": {
    "S": "ScheduleMessage=Started on {year}/{month}/{day} at {hour}:{minute} {timezone}"
  },
"stopped_tags": {
    "S": "ScheduleMessage=Stopped on {year}/{month}/{day} at {hour}:{minute} {timezone}"
  }
```

There are no exceptions logged in the CloudWatch Log output for the tagging process, but they are simply not appearing against the instance. I was expecting to see a tag with the key name of "ScheduleMessage" and a value of "Started on 2018......".

CliveJL:
Looking at the CloudTrail logs I see no evidence of an attempt to call the "CreateTags" API either. It looks like the IF statement in the code (ec2_service.py : if len(instances_starting) > 0:) is not being entered.

anosh237:
Same issue for me as well."Automated Tagging" feature doesn't seem to be working on EC2 instances. I havent tried it yet for RDS.

arieleeuw:
This might have to do with a delay between the instance getting in a "starting" state after the start call has been made. I will look into this

arieleeuw:
We're testing an update that should address the issue if there is a delay between the start/stop of the instance and the instance getting into the desired state.
Please also note that if the instance is already in the desired state (manually started/stopped) at the start/end of a period the scheduler will detect this and will not take any action and by design, will not tag the instances

shsenior:
Closing due to time elapsed since last correspondence. Please re-open this if you are still having an issue.

PR #19, opened 2018-04-26
Fix for issue #2

welderpb:
#2


arieleeuw:
Hi thanks for your PR, good work and thanks for diving deep into this. We've fixed this issue (besides this there was also a timing issue regarding the new state becoming available of started and stopped instances in some regions). We've tested the changes and the update will be available very soon

PR #24, opened 2018-05-19
Update instance_scheduler.py

robertcurcio:
Makes this file govcloud compatible


robertcurcio:
The purpose is allowing the user the option to convert their instance_scheduler.py for use with GovCloud.

gockle:
@robertcurcio The solution version v1.4.0 is no available in GovCloud. This PR can be closed.

PR #25, opened 2018-05-19
Create instance-scheduler.govcloud.template

robertcurcio:
add your bucket name to line 1134


FrankFioretti:
Thank you for your work. Are you planning to update the new version of the scheduler which supports Aurora to support GovCloud?

rakshb:
@robertcurcio We enabled GovCloud support in V1.4.0

PR #30, opened 2018-06-01
Add support for instances in auto scaling groups

DanielRedOak:
We've been using this implementation for awhile with success so I figured I'd submit a PR for consideration.

- Most important to this implementation is having an appropriately set minimum in the ASG. Since Standby mode decrements, increments, and takes nodes out of service, the min needs to be set to allow this action or instances will not be stopped. On our QA instances, we set this to 0 so that all instances can be Stopped when not in use but you can choose any appropriate # and it will be respected.
- When doing discovery on instances, if they belong to an ASG the name of the ASG is added to the instances attributes
- When doing calls to the ASG API, instances are grouped according to their ASG so that we can do batches to limit API calls
- If an instance is scheduled to be stopped and is in an ASG, first we call the API to put it in StandBy mode. If it fails, the instance is removed from the stopping_instances so that the scheduler will not stop it while it is still in the ASG.
- If an instances is scheduled to be started and is in an ASG, it will start first on a scheduler run and move to a 'Standby state' in Dynamo. On the next run it will attempt to put the instance InService. If the call to put the instance InService fails, a warning is logged but no other action is taken since we were able to start the instance.
- The batching generator was modified to allow sending back of instances for individual retries
- Batching is used for Enter/ExitStandby calls. If the batch fails, the instances are sent to the batching generator and retried individually so we can try our best to Enter/Exit Standby mode for instances and Start/Stop them

### Flow diagram

aws instance scheduler flow 1


arieleeuw:
Great piece of work. We will look into the ASG functionality and see if we could add this to the base-line (balance between #of customer requests, performance impact, available time etc..)

For security reasons I recommend to keep the step to grant access to KMS keys a "manual" one giving permissions to any key could involve a security risk (and in many organisations access to KMS keys is managed and audited by a separate team)

Again good work !

DanielRedOak:
Just checking in on this PR. We've been running this in our environments for the past few months without any issues, would love to see it merged.

mludvig:
Is this available in the current release? We need ASG scaling to 0 after business hours as well. Thanks!

Akin2Silver:
Did this functionality end up getting added?

rdkls:
This feature is something we need to take the Instance Scheduler from a neat toy, to something we can apply to serious deployments; majority of which should be using ASGs (even size 1 per best prac).

For this reason I'm about to POC @DanielRedOak SRE-8675 branch (thank you Daniel, and for reporting back that it's working ok) to roll out for our tenancies - but it really sucks not being able to use the official release.

DanielRedOak:
We're still using our version of the scheduler in production and it's saving us a lot on environment costs as we're able to shut things down when not in use. Fixed a few bugs over time and am debating rebuilding it in Ruby since I'm working at a mostly Ruby shop atm. So far we started with ECS support and are using both this version and our new Ruby ECS version in tandem. That version isn't public yet, but we might be open to open sourcing it.

maykays:
@DanielRedOak - Thank your for your submission. Auto-scaling is a native service feature so we do not have plans to implement this feature at the present time.

Issue #38, opened 2018-08-06
Problem starting all instances in a schedule/period


dghadge:
Hello -- Thanks for this great tool. I have around 5 instances that use a single schedule but multiple periods. The periods differ only by the stop time not start time. All those periods are tied to that single schedule.  It stops all the instances in the schedule however while starting it starts only 3 out of 5. These instances are not in ASG or anything special. 
Spent a lot of time debugging - any help is appreciated.

arieleeuw:
Hi,
Can you check if these instances have encrypted volumes?
In that case allow the kms:CreateGrant action to the Scheduler and cross account roles.
Hopes this helps
Arie


Sent from my iPhone

On 6 Aug 2018, at 15:02, Dnyanesh Ghadge <notifications@github.com<mailto:notifications@github.com>> wrote:


Hello -- Thanks for this great tool. I have around 5 instances that use a single schedule but multiple periods. All those periods are tied to that single schedule. It stops all the instances in the schedule however while starting it starts only 3 out of 5. These instances are not in ASG or anything special.
Spent a lot of time debugging - any help is appreciated.

dghadge:
Ughh; i should have know this. Yes they have encrypted volumes. Thank You for quick response and awesome support !!

arieleeuw:
No problem, took me a couple of hours as well ;-)
I've asked our tech writer to update the documentation


Sent from my iPhone

On 6 Aug 2018, at 15:36, Dnyanesh Ghadge <notifications@github.com<mailto:notifications@github.com>> wrote:


Ughh; i should have know this. Yes they have encrypted volumes. Thank You for quick response and awesome support !!

shsenior:
Resolving this issue as the documentation was updated to reflect this.

PR #41, opened 2018-09-01
Add KMS CreateGrant to IAM policy for encrypted EBS support

duckfez:
Adds support for starting EC2 instances with encrypted EBS volumes attached.  Should be a fix for issue #38

### DIFF
--- a/source/cloudformation/instance-scheduler-remote.template
+++ b/source/cloudformation/instance-scheduler-remote.template
@@ -94,7 +94,34 @@
                             ]
                         }]
                     }
-                }]
+                },
+                {
+                    "Effect": "Allow",
+                     "Action": [
+                         "kms:CreateGrant"
+                     ],
+                     "Resource": [
+                         {
+                             "Fn::Join": [
+                                 ":",
+                                     [
+                                         "arn:aws:kms",
+                                         {
+                                             "Ref": "AWS::Region"
+                                         },
+                                         {
+                                             "Ref": "AWS::AccountId"
+                                         },
+                                         "key/*"
+                                    ]
+                              ]
+                         } ],
+                     "Condition": {
+                         "Bool": {
+                             "kms:GrantIsForAWSResource": true
+                         }
+                     }
+                 }]
             }
         }
     },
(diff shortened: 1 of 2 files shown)


arieleeuw:
Thanks for this pull request. We recently added this step in the documentation, instead of putting it in the template. The reason for this is that giving a kms:GreateGrant for all encryption key, would be giving the scheduler role more permissions than it really needs. We recommend to add the permission only for fully specified arn's for keys that are used for the encryption of the volumes of scheduled instances.
Thanks,
Arie

duckfez:
Yes, this was a problem I wrestled with.  In our case, that is almost all of our KMS keys.  Our Terraform configuration for our services generates a KMS key for every group of systems dedicated to a given customer.  I thought about adding a parameter to the template to allow the user to provide a list of key ARNs to grant, but I'll be honest my cloudformation skills aren't that high.

Is this something we could solve via tagging?   As in, "if you tag a KMS key with a scheduler tag, we will add the kms:createGrant privileges as needed to the scheduler's IAM role"

prateek121:
I have come across this too. But may I know how is it related to starting of instances with encrypted volumes which requires createGrant permission ?

duckfez:
@prateek121 long story short, for EC2 to be able to start an instance that accesses encrypted EBS volumes, the EC2 hardware/hypervisor must have access to use the encryption keys for the volumes in question.  Typically this means the principal requesting that the instance be started must have the authority to offer a grant of key access to the EC2 service.  Because instance scheduler runs a lambda function that has an IAM role associated with it, that role needs access to createGrant to the necessary keys.

prateek121:
I see its not stopping the instances with encrypted volumes after giving **kms:createGrant permission**
However, Before giving the permission I see at-least it did stop the instances and started a few instances with encrypted volumes.

prateek121:
Thanks @arieleeuw @duckfez .
Sorry, I see the documents updated.
I had an earlier document downloaded and viewed in chrome from local. 

What I did was apart from createGrant(which was not enough or wasn't working) I have given the permission for list* and describe* and it worked as expected in all accounts.

georgebearden:
Closing this PR but please let us know if something still needs to be addressed.

Issue #71, opened 2019-02-13
Support 'Name' property in Custom Resource


aparker-cityworks:
I have been able to create new Schedules using the Custom Resource in a separate stack.  The name of the Schedule is created by concatenating the Stack name with the name of the resource.  This makes the name rather long and not very good for using as a tag value.

I assumed that I could define the Name property on the custom resource to over ride that behavior. However I got this error when I tried that.
```
ERROR : Name is not a valid property for a schedule, valid schedule properties are Description, Enforced, Metrics, OverrideStatus, Overwrite, Periods, StopNewInstances, Timezone, UseMaintenaceWindow, ServiceToken, Timeout
```
Could you please add the ability to override the name of the schedule by specifying the Name property on the Custom Resource?  I see that you are getting ready to release a new version soon.  Please include this in the upcoming release!

arieleeuw:
Hi,
In the upcoming release the custom resource has a property named "NoStackPrefix" that suppresses the stack name prefix. The schedule will take the name of the resource without the prefix if the value is set to True
Hope this works for you

aparker-cityworks:
That is better than the previous behavior, but still a bit limited.  For example, it's not possible to use a dash (-) in the name of a CloudFormation resource.

I still think a better solution would be to support the Name parameter on the Custom Resource.

aparker-cityworks:
From the CloudFormation documentation https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resources-section-structure.html

> The logical ID must be alphanumeric (A-Za-z0-9) and unique within the template.

That greatly limits what you can name a Schedule when using the Custom Resource.  If one could specify a Name property, then there would be no limitation when using the Custom Resource.  Is it possible for this to be implemented before the pending release?

arieleeuw:
The next update will have a Name property to explicitly set the name of the schedule and a NoStackPrefix to disable the name of the Stack as a prefix

aparker-cityworks:
That will be perfect.  Do you have an estimation of release time frame?

shsenior:
Fixed in v1.2.0

PR #73, opened 2019-02-14
Allow the Name parameter to be sent for a schedule definition.

aparker-cityworks:
Allow the Name parameter to be sent for a schedule definition.  Use the Name property for the name of the Schedule, otherwise fall back to the previous functionality.

This addresses #71.  When I tested this it successfully created a schedule with the name passed in by the Name property.  From what I could tell this is all that needed to be changed, though I could be missing something.  I know you are getting ready to release a new version soon.  Hopefully this small change can be incorporated.

### DIFF
--- a/source/code/requesthandlers/schedule_resource_handler.py
+++ b/source/code/requesthandlers/schedule_resource_handler.py
@@ -40,6 +40,7 @@ PROP_INSTANCE_TYPE = "InstanceType"
 PROP_METRICS = "Metrics"
 PROP_MONTH_DAYS = "MonthDays"
 PROP_MONTHS = "Months"
+PROP_NAME = "Name"
 PROP_OVERRIDE_STATUS = "OverrideStatus"
 PROP_OVERWRITE = "Overwrite"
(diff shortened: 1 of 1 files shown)


arieleeuw:
Thanks for this suggestion.

In the next update the "Name" property will be supported to overwrite the logical resource name as the name of the schedule. There is also a "NoStackPrefix" property that can be set to True in order to get rid of the name of the stack as the prefix of the schedule name.


```
PROP_NAME = "Name"

PROP_NO_STACK_PREFIX = "NoStackPrefix"


   @property
    def _schedule_resource_name(self):
        name = self.resource_properties.get(PROP_NAME,None)
        if name is None:
            name = self.logical_resource_id
        if str(self.resource_properties.get(PROP_NO_STACK_PREFIX, "False")).lower() == "true":
            return name
        return "{}-{}".format(self.stack_name, name)
```

aparker-cityworks:
This is great!  Thanks for getting this in before the update.

aparker-cityworks:
Fixed in v1.2.0

PR #86, opened 2019-04-19
Use on demand provisioning for Dynamo instead of autoscaling

neg3ntropy:
*Description of changes:*

DynamoDb on demand capacity is a much better fit for this use case where most of the time there is zero activity and the rest is short bursts. 

I am pretty sure this saves money over paying for provisioned throughput, bringing the cost of the solution well into the free tiers for normal cases.

As you see from the diff it also saves ~340 lines of cloudfromation.

I have been running like this for a couple of days with no issues.

--


arieleeuw:
Hi,

Fully agree, it saves many lines of code and in most common situation costs as well. In our upcoming 1.2 release we already made the changes to switch from auto scaling to on demand provisioning. This version is in preview and we hope to get it soon released through our internal process.

neg3ntropy:
Hi @arieleeuw,
Thanks for sharing, it would be nice if your management understood that it is not cool to do open source development behind closed doors, since it leads to duplicate effort for the community.

If anybody is going to install this before the new release, I would still recommend that they pick up this patch, since it probably saves some money.

neg3ntropy:
1.2 has this

Issue #95, opened 2019-06-13
iam permission issues on remote template


sreeramsanjay:
Error : 2019-06-13 - 00:31:04.734 - ERROR   : Error stopping rds instance RDS:rds-instance (Oracle EE TEST), (An error occurred (AccessDenied) when calling the StopDBInstance operation: User: arn:aws:sts::xxxxxxxx:assumed-role/instance-scheduler-t-EC2SchedulerCrossAccount-BG948RID8JGI/rds-scheduler-xxxxxxxxx is not authorized to perform: rds:StopDBInstance on resource: arn:aws:rds:us-east-1:xxxxxxxxx:snapshot:rds-ins)

arieleeuw:
Did you create or update a new cross account role with the new template for this version? The cross account role needs additional permissions.
https://docs.aws.amazon.com/solutions/latest/instance-scheduler/templates.html

On 13 Jun 2019, at 02:45, sreeramsanjay <notifications@github.com<mailto:notifications@github.com>> wrote:


Error : 2019-06-13 - 00:31:04.734 - ERROR : Error stopping rds instance RDS:rds-instance (Oracle EE TEST), (An error occurred (AccessDenied) when calling the StopDBInstance operation: User: arn:aws:sts::xxxxxxxx:assumed-role/regn-instance-scheduler-t-EC2SchedulerCrossAccount-BG948RID8JGI/rds-scheduler-xxxxxxxxx is not authorized to perform: rds:StopDBInstance on resource: arn:aws:rds:us-east-1:xxxxxxxxx:snapshot:rds-ins)

jolcese:
I'm having the same problem.
And the weird thing is that is trying to do a StopDBInstance on a snaphost (see arn) which makes no sense.

Another problem is that the role is missing                   
                  - 'rds:AddTagsToResource'
                  - 'rds:RemoveTagsFromResource'
for the Aurora clusters.

georgematthew:
I am also seeing this problem in the cross-account role for version v1.2.0.

skurlow:
I'm having a similar issue with v1.2.0. What is the fix?

[...]

PR #107, opened 2019-07-12
Update instance scheduler cross account role

jolcese:
*Issue #95

Description of changes:

When using snapshots, both the DescribeDBSnaphot and StopDBInstance need to be applied to the snapshot resource also and not only to the rds resource


georgebearden:
Thank you Jose - We are working on the next update of this solution now and will be incorporating this PR.

georgebearden:
This has been incorporated into the latest release.

PR #116, opened 2019-08-14
Fix rds no snapshot

neg3ntropy:
*Description of changes:*
Badly indented block was causing a NameError in case snapshot are disabled for RDS.
First commit fixes whitespace in all files, it was just easier to fix once rather than fight the editor.


georgebearden:
This has been incorporated into the latest release.

PR #117, opened 2019-08-14
Build improvements

neg3ntropy:
*Description of changes:*
This allows custom deployments of the project by:
- allowing different acls for artifacts on S3
- not requiring a mirror bucket in the region, which is obviously something only AWS solutions has.

I don't think this should break anything in your current workflow, apart for requiring jinja to have proper templating. Pipenv config was added, so that should be a plus as well.

I can deploy with:
```
$ make bucket=xyz-instancescheduler acl=private acl_remote=public-read s3_mirrors=false deploy
```


CrypticCabub:
stale -- solution now uses poetry/tox for build management

PR #119, opened 2019-08-20
Verison 1.2.1

YauheniTsyrlin:
Add possibility to schedule instances in autoscaling groups by detaching
and attaching instances.
Fix RDS disabled snapshots.
Add privileges to IAM.

*Issue #, if available:*

*Description of changes:*


hearde:
Thanks for this contribution, and sorry for the long period without a response. While we want to add support for ASG, we will likely not do so by attaching and detaching instances. The scaling should be managed at the group level.

It looks like the fix for RDS snapshots has already been integrated.

PR #170, opened 2020-06-10
Fix None value for instance type when creating period/schedule

floriankauschke:
Due to instancating a new period, the call for the instance type in line 163 via period.get(PROP_INSTANCE_TYPE, None) would be performed on the newly created period, where no instance type can be found. Therefor the result would be always empty.
The resolution for this is to call this before the period is overwritten.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,13 @@ source/deploy
 source/code/sample_events
 .vscode
 __pycache__
+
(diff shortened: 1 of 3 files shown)


praveenis:
Thanks @floriankauschke for the request. We have added this to our backlog and will update in a future release

PR #182, opened 2020-07-08
GovCloud Intergration

NathanGilbert2:
Templates updated for use in both validated standard AWS regions and AWS GovCloud (US) regions via the AWS::Partition parameter. Successful CloudFormation build in both partitions (Regions built in: us-east-1, us-gov-west-1, us-gov-east-1) Used my own buckets for GovCloud build. Attention from solution owners would be needed on the back end since the solution resources are not available in buckets "solutions-us-gov-west-1" and "solutions-us-gov-east-1".

*Issue  11, 105 

*Description of changes:*
Modified principal and policy ARNs to dynamically detect the partition they are in using [AWS::Partition](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html#cfn-pseudo-param-partition) pseudo parameter.

### DIFF
--- a/source/cloudformation/instance-scheduler-remote.template
+++ b/source/cloudformation/instance-scheduler-remote.template
@@ -11,14 +11,16 @@
     },
     "Metadata": {
         "AWS::CloudFormation::Interface": {
-            "ParameterGroups": [{
-                "Label": {
-                    "default": "Account"
-                },
-                "Parameters": [
-                    "InstanceSchedulerAccount"
-                ]
-            }],
+            "ParameterGroups": [
+                {
+                    "Label": {
+                        "default": "Account"
+                    },
+                    "Parameters": [
+                        "InstanceSchedulerAccount"
+                    ]
+                }
+            ],
             "ParameterLabels": {
                 "InstanceSchedulerAccount": {
                     "default": "Primary account"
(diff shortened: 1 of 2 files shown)


rakshb:
@NathanGilbert2 Thanks for the PR. We will add this to our backlog and review it in a future release.

NathanGilbert2:
> @NathanGilbert2 Thanks for the PR. We will add this to our backlog and review it in a future release.


Would like to also see if the solutions buckets in us-iso-east-1 and us-isob-east-1 could include this update as well. I am sure there are those interested in using this in those isolated regions as well.

scottstaffiere:
I am working on GovCloud integration today and found this.  Wanted to thank you for the template - that helps quite a bit and makes it much more portable.  The other suggestion I'd like to make is to fix the __init__ method in the instance_scheduler.py file (around line 91).  

This piece of code throws an error about invalid regions in GovCloud because it assumes the partition is 'aws':
`boto3.Session().get_available_regions(service.service_name)`

It would be nice to automatically detect the partition based on boto3.Session().region_name, or if there is a partition variable set somewhere else.  I ended up doing something like this on my end:
`partition = if 'gov' in boto3.Session().region_name then 'aws-us-gov' else 'aws'
 self._valid_regions = boto3.Session().get_available_regions(service.service_name, partition)`

It would be great though to get into the base set of code so it doesn't need to be overridden on our end.

rakshb:
@NathanGilbert2 We enabled GovCloud support in V1.4.0.

PR #187, opened 2020-07-28
Add exports to template

Wade9320:
*Issue #, if available:*
[Issue 188](https://github.com/awslabs/aws-instance-scheduler/issues/188)

*Description of changes:*

Added exports to the templates so they can be used by custom resource CloudFormation instead of hard-coding the lambda arn: https://docs.aws.amazon.com/solutions/latest/instance-scheduler/appendix-d.html

### DIFF
--- a/deployment/instance-scheduler.template
+++ b/deployment/instance-scheduler.template
@@ -1675,7 +1675,7 @@
             "Value": {
                 "Ref": "ConfigTable"
             },
-            "Description": "Name of the DynomoDB configuration table",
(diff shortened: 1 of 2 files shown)


gockle:
Hi @Wade9320 The solution's scheduler-cli can be used to create/update/remove schedules and is much more flexible it will not require multiple cloudformation templates, as an alternative to consider using [scheduler-cli](https://docs.aws.amazon.com/solutions/latest/instance-scheduler/scheduler-cli.html).

PR #201, opened 2020-09-24
Fix cron syntax in cloudformation template

yatoum:
I recently tried to create a cloudformation stack using the latest template with the Frequency parameter set to 60, but the creation failed due to the error "Parameter ScheduleExpression is not valid."
To fix this issue, I fixed the cron syntax in the cloudformation template.


gockle:
@yatoum Sorry for the issue, we will incorporate this change in a future release.

aiell0:
@gockle bumping this. Is there any reason why a one-line code change takes almost two months to release? It's a bit frustrating that this is a known issue and has been for months, and expecting users to just deal with fixing a 1700 line JSON template would have me looking for another solution quickly if I wasn't fluent in this already. Just food for thought.

gockle:
Hi @aiell0 @yatoum  The cron schedule has been updated in the source code and the template.

Issue #205, opened 2020-10-02
CloudFormation template error (Mapping --> Timeouts)


eduardopuente:
Parameter ScheduleExpression is not valid. (Service: AmazonCloudWatchEvents; Status Code: 400; Error Code: ValidationException

This is because on the defined mappings, there is an error:
`"60": "cron(0/1 * * ? *)"` should be `"60": "cron(0/60 * * ? *)"`

gockle:
@eduardopuente Thanks for providing this information, we will patch is error in a future release.

paultung:
The actually error is here:
[source/lib/aws-instance-scheduler-stack.ts#L615](https://github.com/awslabs/aws-instance-scheduler/blob/107aa722d01c32713ed31a6acbb73b0395527387/source/lib/aws-instance-scheduler-stack.ts#L615)

I think it should be:
`cron(0/60 * * * ? *)` or `cron(0 0/1 * * * ? *)`

FYR~

gockle:
@eduardopuente Issue has been resolved in the latest version of the solution v1.4.0.

PR #210, opened 2020-11-01
fix: Fix event cron job in 60 seconds

paultung:
*Issue* #205 

*Description of changes:*
1. It should be 60 seconds not 1 second.
2. Add back lacked field to Cron expression.


rakshb:
@paultung Thank you for reporting this. This is the same as https://github.com/awslabs/aws-instance-scheduler/pull/201

We will address this in the next release

gockle:
@paultung The cron expression has been updated in the source code.

PR #223, opened 2021-01-08
add Name property to Periods in cloudformation custom resource template

Celestior:
*Issue:*
Specifying period name within Custom Resource ServiceInstanceSchedule is leading the Cloudformation stack update to fail. When not specified in the template, the period name is generated automatically in this format: "{Schedule_name}-period-{period_number}" without giving users the possibility to define a period name. But using scheduler-cli it is possible to define a period name.

*Description of changes:*
Adding the possibility to specify period name within Custom Resource ServiceInstanceSchedule by adding property "Name" to Periods property.


gockle:
The schedule/period names can be modifed and set to custom values using the scheduler-cli or the by simply editing them in the DynamoDB table console.

Issue #222, opened 2020-12-16
Feature Request: Add support for DocumentDB Instances


dbayardAWS:
Given the similarities with RDS's CLI, would it be possible to add support for scheduling DocumentDB instances, too?

gockle:
Hi @dbayardAWS Thank you for reaching out, we will add this to our feature request backlog and try to include it in a future release of the solution.

ajhindle:
Hi, I'd be grateful to see this feature implemented also. 
The request for this has been open since at least 2019 (see linked issue below, it should be open).
This feature is needed because Document DB is expensive to run!
https://github.com/awslabs/aws-instance-scheduler/issues/111

rakshb:
This feature request has been prioritized in our backlog for a future release.

rakshb:
Duplicate of #336

PR #253, opened 2021-05-12
Add DocumentDB support

cdvieu:
Issue #222 

Description of changes:
Adding functionality to support the scheduling of DocDB resources. Updated rds_service to exclude DocDB resources as DocDB was implicitly included due to overlapping API calls.

### DIFF
--- a/source/lambda/configuration/config_admin.py
+++ b/source/lambda/configuration/config_admin.py
@@ -86,7 +86,7 @@ class ConfigAdmin:
     # regex for checking time formats H:MM and HH:MM
     TIME_REGEX = "^([0|1]?[0-9]|2[0-3]):[0-5][0-9]$"
 
-    SUPPORTED_SERVICES = ["ec2", "rds"]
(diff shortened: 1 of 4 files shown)


maykays:
Thank you for the PR. We are unable to support this PR at the moment based on a scaling concern. We are working towards a solution for Q3/4 2021.

brianmjones:
@maykays I have a few questions. What is the scaling concern? Is the scaling concern with the existing codebase or with our code? Are you expecting us to do another PR when the scaling concern is addressed?

gockle:
Hi @brianmjones Scaling concerns are with the solution itself, the current design of instance scheduler solution has some limitation which needs to be addressed, we are currently evaluating a new other design options for the solution, Hence at this point we don't want to include/expand this solution in its current architecture to more service. The PR looks good technically no more changes are required from your end. Let me know if you have any further questions.

cdvieu:
Is the solution for the scaling concerns still scheduled to happen in Q3/Q4? We're looking for an update for planning purposes. Thanks!

Issue #250, opened 2021-05-06
EBS KMS permissions not included in remote role


markyates:
const ec2Permissions = new iam.Policy(this, "Ec2Permissions", {
      statements: [
        new PolicyStatement({
          actions: [
            'ec2:ModifyInstanceAttribute'
            'kms:CreateGrant',
          ],
          effect: Effect.ALLOW,
          resources: [
            cdk.Fn.sub("arn:${AWS::Partition}:ec2:*:${AWS::AccountId}:instance/*")
          ]
        }),
        ec2PolicyAssumeRoleStatement
      ],
      roles: [schedulerRole]
    })

As you can see the github links above contains the main TS templates that generates the CFN yaml/json templates.
==================================================
Please feel free to copy the ec2Permissions code and create a new policy with kms permissions , or you can add the permission to the existing one, I would recommend adding a github issues to allow the team taking such requirement into consideration within the solution.

gockle:
@markyates Sorry about the issue, we will review this and add it to the solution backlog. It will be prioritized for an upcoming release.

rakshb:
Adding a permission to provide scheduler role to have access to all the KMS keys in the account will have unintended security issues, we will review this scenario for the solution, but this PR will not be merged into the repository. I would suggest to narrow down the permission to KMS keys in your account to only the ones that the scheduler really requires.

PR #271, opened 2021-09-23
allow scheduler to work with encrypted resources

elduds:
Fixes Issue #250 

*Description of changes:*
Allows scheduler data operations on all KMS keys (Resource: "*" as ID of CMKs in remote accounts is not known/knowable and could be in a different account to the remote account).


elduds:
not working

PR #314, opened 2022-06-01
fix for changing instance type in period

whoDoneItAgain:
*Issue #, if available:* #315

*Description of changes:*
swap lines 159 and 160 to fix issue where a period configured to change instance size is not reflected in the schedule's period(s)


hearde:
Thank you for the contribution, and sorry for the long delay in response. We will include a fix for this in our next release.

PR #316, opened 2022-06-04
Updated example commands

michalani:
The example commands are all over the place, changed to match a new user trying to use this solution.

*Description of changes:*
changed the order of commands to match new users.


CrypticCabub:
Readme has been updated with new installation requirements for updated organization

PR #345, opened 2023-01-11
Hotfix to update cdk version to v2 and rebuild Lambda Python to update certifi

gockle:
*Issue #, if available:*
* Address issue with python package [certifi](https://security.snyk.io/vuln/SNYK-PYTHON-CERTIFI-3164749)

*Description of changes:*
1. Changes to the components which generate the cloudformation template to utilize cdk v2 components
2. Changes to snapshot test files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.4.2] - 2023-01-11
 ### Fixed
-- Updated the version of certifi to 2022-12-07 to resolve vulnerability https://security.snyk.io/vuln/SNYK-PYTHON-CERTIFI-3164749
+- Upgrade certifi to mitigate [CVE-2022-23491](https://nvd.nist.gov/vuln/detail/CVE-2022-23491).
 - Updated issues in bandit scan.
 - Updated the CDK version 2.x
 
--- a/deployment/build-s3-dist.sh
+++ b/deployment/build-s3-dist.sh
@@ -116,7 +116,7 @@ mv version.py version.py.org
 sed "s/%version%/$DIST_VERSION/g" version.py.org > version.py
 
 echo "Install all the python dependencies in the staging directory before packaging"
-pip3 install -U -r $source_dir/lambda/requirements.txt -t $staging_dist_dir/lambda/
+pip install -U -r $source_dir/lambda/requirements.txt -t $staging_dist_dir/lambda/
 
 echo "Build lambda distribution packaging"
 zip -q --recurse-paths ./instance-scheduler.zip version.txt main.py version.py configuration/* requesthandlers/* chardet/* urllib3/* idna/* requests/* schedulers/* util/* boto_retry/* models/* pytz/* certifi/*


.viperlightignore
```diff
@@ -0,0 +1,26 @@
+# Ignore Config used with code.amazon.com
```
hearde, line 1: What's up with these files that should already be there?

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.2] - 2023-01-11
+### Fixed
+- Updated the version of certifi to 2022-12-07 to resolve vulnerability https://security.snyk.io/vuln/SNYK-PYTHON-CERTIFI-3164749
```
hearde, line 9: I put this item under `Security`: https://keepachangelog.com/en/1.0.0/
I also use the CVE id rather than going through snyk: `- Upgrade certifi to mitigate CVE-2022-23491`

buildspec.yml
```diff
@@ -0,0 +1,39 @@
+version: 0.2
+
+phases:
+  install:
+    runtime-versions:
```
hearde, line 5: Not necessary now, but you can also fix your node version to 16. It's the only version available in the 6.0 build containers, but you'll be alerted when/if that changes or if you switch pipeline versions.

buildspec.yml
```diff
@@ -0,0 +1,39 @@
...
+      python: 3.10    
+    commands:
+      - pip install --upgrade pip
```
hearde, line 8: I'd also upgrade setuptools

buildspec.yml
```diff
@@ -0,0 +1,39 @@
...
+      - pip install -r source/lambda/requirements.txt -t source/lambda/
```
hearde, line 9: I don't like putting things like this in the buildspec because it makes the local and pipeline builds different. I'd rather put it in build-s3-dist if it's really necessary, and clean up afterward.

buildspec.yml
```diff
@@ -0,0 +1,39 @@
...
+  pre_build:
+    commands:
+      - echo "Installing dependencies and executing unit tests - `pwd`"
+      - cd source
+      - chmod +x ./run-all-tests.sh && ./run-all-tests.sh
```
hearde, line 14: `chmod` should not be necessary, multiple instances

buildspec.yml
```diff
@@ -0,0 +1,39 @@
...
+phases:
+  install:
+    runtime-versions:
+      python: 3.10    
+    commands:
+      - pip install --upgrade pip
+      - pip install -r source/lambda/requirements.txt -t source/lambda/
+  pre_build:
+    commands:
+      - echo "Installing dependencies and executing unit tests - `pwd`"
```
hearde, line 12: I don't like echoing commands - they can be out of date

buildspec.yml
```diff
@@ -0,0 +1,39 @@
...
+      - cd ../deployment
+      - echo "Starting build `date` in `pwd`"
+      - chmod +x ./build-s3-dist.sh && ./build-s3-dist.sh $DIST_OUTPUT_BUCKET $SOLUTION_NAME $VERSION
+      - echo "Build completed `date`" 
+      - echo "Starting open-source-dist `date` in `pwd`" 
+      - chmod +x ./build-open-source-dist.sh && ./build-open-source-dist.sh $SOLUTION_NAME
+      - echo "Open Source Dist completed `date`" 
+
+  post_build:
+    commands:
```
hearde, line 29: This whole section should be unnecessary in v6 pipelines

deployment/build-s3-dist.sh
```diff
@@ -13,7 +13,7 @@
 #
 
 # Important: CDK global version number
-cdk_version=1.96.0
+cdk_version=2.59.0
```
hearde, line 16: I don't think this is necessary. You can either trust your package-lock/npm-shrinkwrap or fix the exact version in the package file. I'm not sure that installing a specific version over what might already be there is guaranteed to have a good outcome.

deployment/build-s3-dist.sh
```diff
@@ -116,7 +116,7 @@ mv version.py version.py.org
 sed "s/%version%/$DIST_VERSION/g" version.py.org > version.py
 
 echo "Install all the python dependencies in the staging directory before packaging"
-pip install -U -r $source_dir/lambda/requirements.txt -t $staging_dist_dir/lambda/
+pip3 install -U -r $source_dir/lambda/requirements.txt -t $staging_dist_dir/lambda/
```
hearde, line 119: Prefer to do this in a venv, but it shouldn't matter for something like this

source/lambda/testing_requirements.txt
```diff
@@ -1,7 +1,7 @@
 botocore>=1.12.99
```
hearde, line 1: Why not lock all package versions?

source/lambda/testing_requirements.txt
```diff
@@ -1,7 +1,7 @@
...
 boto3>=1.9.99
```
hearde, line 2: Technically boto3 and botocore are held back in lambda runtimes, I try to match, but it probably doesn't make a difference: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

source/package.json
```diff
@@ -12,30 +12,19 @@
...
-    "@aws-cdk/assert": "1.96.0",
-    "@types/jest": "^25.2.1",
-    "@types/node": "10.17.5",
-    "aws-cdk": "^1.96.0",
-    "jest": "^26.6.3",
-    "ts-jest": "^26.5.3",
-    "ts-node": "^8.1.0",
+    "@aws-cdk/assert": "^2.59.0",
+    "@types/jest": "^26.0.10",
+    "@types/node": "10.17.27",
```
hearde, line 17: Any particular reason for the versions that are locked in this file?

source/test/aws-instance-scheduler-remote-stack.test.ts
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env node
 /*****************************************************************************
```
hearde, line 1: Standardize on the short copyright header at some point

hearde (approved):

.viperlightrc
```diff
@@ -0,0 +1,4 @@
+{
+  "all": true,
```
tbelmega, line 2: Are you aware the Viperlight config will only be used in case of a local scan?
Our builder pipeline ignores .viperlightrc and always uses the default settings.

source/lambda/testing_requirements.txt
tbelmega: I tend to use `botocore~=1.12.99` so that new patch versions of libraries are automatically used without the need to change the library version in (testing-)requirements.txt.
If I fix the version with "==", every published patch version release will cause the Viperlight prebuild scan to fail, costing extra cycles. The annoyance outweighs the benefits imho.

PR #346, opened 2023-01-11
Adding changes to remove additional cloudformation parameter Bootstrap version

gockle:
*Issue #, if available:*

Not available.

*Description of changes:*

Remove additional CloudFormation parameter `Bootstrap version`

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -6,7 +6,6 @@ phases:
       python: 3.10    
     commands:
       - pip install --upgrade pip
-      - pip install -r source/lambda/requirements.txt -t source/lambda/
   pre_build:
     commands:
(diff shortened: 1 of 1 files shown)


source/bin/aws-instance-scheduler.ts
```diff
@@ -28,6 +28,9 @@ const SOLUTION_PROVIDER = 'AWS Solution Development';
 const app = new cdk.App();
 
 new AwsInstanceSchedulerStack(app, 'aws-instance-scheduler', {
+    synthesizer: new cdk.DefaultStackSynthesizer({
```
hearde, line 31: Should this rule also be applied to the test stacks? Shouldn't affect tests, but no reason for them to be different.

hearde (approved):

PR #347, opened 2023-01-11
Chardet is missing dependency for requests, adding it to the requirements.txt

gockle:
*Issue #, if available:*

Not available

*Description of changes:*

Chardet is missing dependency for requests, adding it to the requirements.txt

### DIFF
--- a/source/lambda/requirements.txt
+++ b/source/lambda/requirements.txt
@@ -1,2 +1,3 @@
 pytz
 requests
+chardet


source/lambda/requirements.txt
```diff
@@ -1,2 +1,3 @@
 pytz
-requests
\ No newline at end of file
+requests
+chardet
```
CrypticCabub: Should we specify specific versions here?

CrypticCabub: should we be specifying specific versions in requirements.txt?

Issue #307, opened 2022-03-23
Instance scheduler lambda concurrency count is too high


anuj1366:
Hi Team
I am running Instance scheduler solution every hour for 600 account for single region for both services ec2 and rds.
Lambda timeout is set to 10 minutes and memory is set to 320 mb.
It means every hour count of lambda invocation will be = 600 *1 *2= 1200
I have run this solution for one month and that time lambda concurrency count was 70-80 and it takes around 4 minute to comlete the scheduling of ec2 and rds instances in all the 600 account.
In this case lambda average duration was under 10 sec

But now from last few days lambda concurrency count increased to 1000 and it takes around 1 minute to complete the scheduling of ec2 and rds instances in all the 600 account.
In this case lambda average duration increased to 200 seconds because of that solution cost has been increased.

Could you please let me know what could be the reason that has increased the lambda concurrency count without changing anything in solution.
Do let me know if you need more information.

anuj1366:
I have noticed logger.flush() function is taking more time in this issue.
self._logger.flush() code is present inside finally block in handle_request() method which is present in file "scheduler_request_handler.py".
whole try block inside handle_request() method is taking maximum 2 second but flush function is taking around 30sec -60 sec. because of that Instance scheduler lambda average duration is high which increase the lambda cost as well.

gockle:
Hi @anuj1366 

Based on the information provided there multiple options that can be explored,

1. Has anything changed in the account where the solution is hosted? example are there other Lambda functions being used in the same account? or Has there been any usage increase in the number of CloudWatch API calls being made in that account?
2. The `logger.flush()` is there to make sure what ever log statements are processed to the CloudWatch logs, you can if chose to comment it, and then if in the future you encounter any issue uncomment it to retrieve log events. Let us know if you face any additional issue when commenting the `logger.flush()` line in finally.

anuj1366:
@gockle 
1. We have not changed anything in the account where the solution is hosted.
2. We have commented the code inside flush() function and after that solution is working perfectly fine without any issue.

I have analyze the issue in more details and figure out below line is causing the delay in solution which is present inside flush() method in logger.py file.

**resp = self.client.describe_log_streams_with_retries(logGroupName=self._loggroup, logStreamNamePrefix=self._logstream)**

Currently there are 3000 log stream inside a log group and validity of log is 1 months. 
I am not sure why suddenly above api call has started taking more time.

[...]

PR #353, opened 2023-01-19
Replace boto_retry mechanism with boto3 standard retry mechanism

gockle:
*Issue #, if available:*
#307 

*Description of changes:*
1. Remove the boto_retry mechanism and replace with [boto3 standard retry mode](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html)
2. Remove models loaded into the lambda and use the default lambda service.json

### DIFF
--- a/source/lib/aws-instance-scheduler-stack.ts
+++ b/source/lib/aws-instance-scheduler-stack.ts
@@ -906,7 +906,7 @@ export class AwsInstanceSchedulerStack extends cdk.Stack {
     })
 
     const customServiceCfn = customService.node.defaultChild as cdk.CfnCustomResource
-    customServiceCfn.addDependsOn(schedulerLogGroup_ref)
+    customServiceCfn.addDependency(schedulerLogGroup_ref)
 
     //Instance scheduler Cloudformation Output references.
     new cdk.CfnOutput(this, 'AccountId', {
@@ -1083,10 +1083,10 @@ export class AwsInstanceSchedulerStack extends cdk.Stack {
 
     //Adding the EC2 and scheduling policy dependencies to the lambda. 
     const lambdaFunction = lambdaToDynamoDb.lambdaFunction.node.defaultChild as lambda.CfnFunction
-    lambdaFunction.addDependsOn(ec2DynamoDBPolicy.node.defaultChild as iam.CfnPolicy)
-    lambdaFunction.addDependsOn(ec2Permissions.node.defaultChild as iam.CfnPolicy)
-    lambdaFunction.addDependsOn(schedulerPolicy.node.defaultChild as iam.CfnPolicy)
-    lambdaFunction.addDependsOn(schedulerRDSPolicy.node.defaultChild as iam.CfnPolicy)
+    lambdaFunction.addDependency(ec2DynamoDBPolicy.node.defaultChild as iam.CfnPolicy)
+    lambdaFunction.addDependency(ec2Permissions.node.defaultChild as iam.CfnPolicy)
+    lambdaFunction.addDependency(schedulerPolicy.node.defaultChild as iam.CfnPolicy)
+    lambdaFunction.addDependency(schedulerRDSPolicy.node.defaultChild as iam.CfnPolicy)
     lambdaFunction.cfnOptions.metadata = {
       "cfn_nag": {
         "rules_to_suppress": [


source/lambda/util/boto_utils.py
```diff
@@ -0,0 +1,23 @@
...
+#                                                                                                                    #
+#  or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES #
+#  OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions    #
+#  and limitations under the License.                                                                                #
+######################################################################################################################
+
+import boto
+import botocore
+
+def get_config():
```
hearde, line 17: Why is this here and in init?

source/lambda/util/boto_utils.py
```diff
@@ -0,0 +1,23 @@
...
+    user_agent_extra_string = os.getenv("USER_AGENT_EXTRA", "AwsSolution/SO0030/v1.5.0")
+    solution_config = {"user_agent_extra": user_agent_extra_string, "retries": {'max_attempts': 5, 'mode': 'standard'}}
+    return botocore.config.Config(**solution_config) 
+
+# def get_client(service, context=None, session=None):
```
hearde, line 22: Remove

source/lambda/util/boto_utils.py
```diff
@@ -0,0 +1,23 @@
+######################################################################################################################
```
hearde, line 1: Use the short header

source/lib/aws-instance-scheduler-stack.ts
```diff
@@ -860,7 +860,7 @@ export class AwsInstanceSchedulerStack extends cdk.Stack {
       ],
       effect: Effect.ALLOW,
       resources: [
-        cdk.Fn.sub("arn:${AWS::Partition}:ssm:${AWS::Region}:${AWS::AccountId}:parameter/Solutions/instance-scheduler/UUID/*")
+        cdk.Fn.sub("arn:${AWS::Partition}:ssm:${AWS::Region}:${AWS::AccountId}:parameter/Solutions/aws-instance-scheduler/UUID/*")
```
hearde, line 863: This seems unrelated, what's the purpose?

source/lambda/boto_retry/__init__.py
```diff
@@ -11,136 +11,20 @@
...
-                            method_suffix=DEFAULT_SUFFIX):
+def get_client(service_name, region=None, session=None):
     """
     Creates a bot3 client for the specified service name and region. The return client will have additional method for the
     specified methods that are wrapped with the logic of the specified wait strategy or the default strategy for that service.
     The method names must be valid for the boto3 service client. The name of the added functions is the name of the original
     function plus the (default) value of method_suffix parameter
     :param service_name: Name of the service
-    :param methods: List of methods for which a new method will be added to the client wrapped in retry logic
     :param context: Lambda execution context
```
hearde, line 25: Looks like this param is gone too

source/lambda/boto_retry/__init__.py
```diff
@@ -11,136 +11,20 @@
...
-    else:
-        retry_class = AwsApiServiceRetry
-    return retry_class
-
-
-def get_client_with_retries(service_name, methods, context=None, region=None, session=None, wait_strategy=None,
-                            method_suffix=DEFAULT_SUFFIX):
+def get_client(service_name, region=None, session=None):
     """
     Creates a bot3 client for the specified service name and region. The return client will have additional method for the
```
hearde, line 20: Should be updated - client with standard retry policy

hearde (approved):
When I commented "Should be updated - client with standard retry policy" I was just referring to the docstring for the `get_client` function. I think it's fine to leave it as `get_client`.

It looks like when you select a single line for comment in a gh pr, that line is the last line that appears above the comment in the summary view, and gh seems to choose how much context to give it, so looking at the summary it looks like I was commenting on the function name. I'll try to be more clear in the future, sorry about that.

source/lambda/configuration/config_admin.py
```diff
@@ -108,8 +108,7 @@ def __init__(self, logger, context):
         :param context: Lambda context
         """
         self._table_name = self.table_name
-        self._table = boto3.resource("dynamodb").Table(self._table_name)
-        add_retry_methods_to_resource(self._table, ["scan", "get_item", "put_item", "delete_item"], context=context)
+        self._table = DynamoDBUtils.get_dynamodb_table_resource_ref(self._table_name)
```
CrypticCabub, line 111: I don't see any reference to setting retries in this file, is standard retry the default now for the dynamo interface? (legacy is the default for clients - [doc](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html))

source/lambda/main.py
```diff
@@ -30,21 +30,6 @@
...
-    if aws_data_path is not None:
-        aws_data_path = ":".join([aws_data_path, models])
-    else:
-        aws_data_path = models
-    os.environ["AWS_DATA_PATH"] = aws_data_path
-
-
-load_models()
-
-
```
CrypticCabub, line 47: I presume this has been confirmed as an unused method? It is not clear to me how it pertains to the old retry logic

source/lambda/schedulers/ec2_service.py
```diff
@@ -74,8 +75,7 @@ class Ec2Service:
     EC2_STOPPING_STATES = {EC2_STATE_SHUTTING_DOWN, EC2_STATE_STOPPING, EC2_STATE_STOPPED}
     EC2_STARTING_STATES = {EC2_STATE_PENDING, EC2_STATE_RUNNING}
 
-    dynamodb = boto3.resource('dynamodb')
-    maintenance_table = dynamodb.Table(os.environ['MAINTENANCE_WINDOW_TABLE'])
+    maintenance_table = DynamoDBUtils.get_dynamodb_table_resource_ref(os.environ['MAINTENANCE_WINDOW_TABLE'])
```
CrypticCabub, line 78: we should extract these env variables into a centralized enum registry of some kind, i dislike the magic strings that don't get tested until runtime -- adding as backlog item

CrypticCabub (commented):
I don't see any tests that confirm that the retry logic is actually being correctly applied. This may be an in-built feature of the boto library now but I wonder if it's possible to write a test (maybe with moto?) that would confirm that the client_with_retries is actually attempting retries as expected

CrypticCabub:
> @CrypticCabub testing default boto3 functionality is out of scope for this user story and solution, we rely on boto3 to perform as the their API reference documentation so that we rely on not just retries, but also handling security, HTTP request response.

Implementation details of boto3 are indeed out of scope, but whether or not retries are attempted at all (i.e. is the configuration being correctly picked up) is still something I would like a unit test for

CrypticCabub:
let's include it as a backlog item for the final 1.5 release since we're working on fixing the testing anyway

CrypticCabub (approved):

PR #357, opened 2023-01-25
Configure Tox & SonarQube

CrypticCabub:
*Description of changes:*
-Configure Tox to run all testing for CDK and Lambda
-Setup sonarqube to receive coverage reports during internal pipeline execution


CrypticCabub:
> Can you add instructions in the README.md how to build/test the source code.

The README will likely need to be more thoroughly updated when the build process itself is updated, however currently the instructions in the readme will work just fine (they are actually more correct now than they were before)

PR #360, opened 2023-01-26
Feature/pipeline

CrypticCabub:
-Setup Tox to handle testing of CDK and Lambda code 
--tests can be run by running "tox" from the root of the project
-sonarqube now picks up on coverage reports generated by run-unit-tests.sh script in /deployment
-a few minor refactors merged in from changes made back in December that didn't merge with the initial move to github first


source/lambda/pyproject.toml
```diff
@@ -0,0 +1,42 @@
...
+    "constructs>=10.0.0,<11.0.0",
+]
+
+[project.optional-dependencies]
+dev = [
+    "tox",
+    "boto3-stubs-lite[essential]",
+    "boto3-stubs-lite[ssm]",
+    "botocore>=1.12.99",
+    "boto3>=1.9.99",
```
hearde, line 33: Unless you package the sdk in your lambda or layer yourself, you get a specific version: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

I don't see a reason to test locally with a different version.

source/lambda/schedulers/ec2_service.py
```diff
@@ -75,15 +75,15 @@ class Ec2Service:
...
-    maintenance_table = DynamoDBUtils.get_dynamodb_table_resource_ref(os.environ['MAINTENANCE_WINDOW_TABLE'])
-
     def __init__(self):
         self.service_name = "ec2"
         self.allow_resize = True
         self.schedules_with_hibernation = []
         self._ssm_maintenance_windows = None
         self._session = None
         self._logger = None
+        self._dynamodb = boto3.resource('dynamodb')
```
hearde, line 85: Remember that `resource` isn't receiving updates, I'd avoid building more on top of it if at all possible

source/lambda/tests/botomock.py
```diff
@@ -0,0 +1,12 @@
+from unittest import mock
+import os
+
+
+def mock_boto_env_vars():
```
hearde, line 5: Why not an autouse fixture? Then you don't have to call it every time

source/test/aws-instance-scheduler-remote-stack.test.ts
```diff
@@ -13,51 +13,13 @@
  *  permissions and limitations under the License.                            *
  *****************************************************************************/
 
-import { SynthUtils } from '@aws-cdk/assert';
-import * as cdk from 'aws-cdk-lib';
-import instanceSchedulerStack = require('../lib/aws-instance-scheduler-remote-stack');
-import '@aws-cdk/assert/jest';
+ import { SynthUtils } from '@aws-cdk/assert';
```
hearde, line 16: This library is deprecated, consider migrating away: https://github.com/aws-solutions/aws-security-hub-automated-response-and-remediation/pull/87

buildspec.yml
```diff
@@ -34,5 +33,7 @@ phases:
 artifacts:
   files:
       - deployment/**/*
+      - source/**/*
```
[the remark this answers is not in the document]
CrypticCabub: good point, I'll add a note to review this when I come to reviewing the build process

source/lambda/schedulers/ec2_service.py
CrypticCabub: ec2_service (and related scheduler code) is currently highly fragile and not open to easy modification. The purpose of all these changes to the testing/build process is to allow for tests to be written against these files especially so that edits exactly like you mention can be done safely

source/lambda/tests/botomock.py
CrypticCabub: I'm of two minds about it. Automatically taking care of it is nice in that you can't forget about it, but it's also hell to debug if you don't know there's extra magic happening behind the scenes. 

I had also been having trouble getting the magic to behave correctly due to certain boto elements being loaded statically on file import. I have fixed that in all currently affected test files, but this approach was originally born out of needing to be able to mock the env vars before certain files were imported, and now I generally like that it's clear exactly what is being mocked when.

source/test/aws-instance-scheduler-remote-stack.test.ts
CrypticCabub: added to backlog

source/lambda/pyproject.toml
CrypticCabub: you're probably right. the dependencies list here definitely needs to be pruned and consolidated (it's currently split between the pyproject.toml and requirements/testing_requirements .txt files) Cleaning up the dependencies into a single source of truth would be the scope of the upcoming updates to the build process (my goal here was just to tack a working test process onto the existing build process)

buildspec.yml
```diff
@@ -3,20 +3,19 @@ version: 0.2
...
     commands:
       - pip install --upgrade pip
+      - pip install -r source/lambda/requirements.txt -t source/lambda/
   pre_build:
     commands:
       - echo "Installing dependencies and executing unit tests - `pwd`"
-      - cd source
-      - chmod +x ./run-all-tests.sh && ./run-all-tests.sh
+      - cd deployment
+      - chmod +x ./run-unit-tests.sh && ./run-unit-tests.sh
```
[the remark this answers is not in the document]
CrypticCabub: on the pipeline yes, but run locally this can sometimes fail so this just ensures that the script doesn't get tripped up by a stupid permissions issue. (this script will likely die completely in the coming updates as well)

source/lambda/tests/botomock.py
hearde: That import behavior is the thing I was fixing in ASR, it is frustrating. You can also move the import of the file in question to be within a function or test, so that it occurs after mocking happens.

This fixture is a pretty common pattern, so it's not like it's totally unexpected: https://www.learnaws.org/2020/12/01/test-aws-code/

One reason to use an autouse fixture is that when you forget to add that call, and you write your local tests, and you miss a stub, if your local profile has the permissions (mine is frequently on Admin) you get real calls into your account, which is not fun. Less catastrophically, it can also work locally but fail in the pipeline.

buildspec.yml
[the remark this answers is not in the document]
hearde: I'm inclined to believe that if a user unsets the executable bit on a file cloned from a repo, or sets their `umask` so that happens, they don't want that file to execute.

deployment/run-unit-tests.sh
```diff
@@ -1,6 +1,58 @@
 #!/bin/bash
-cd ..
-echo 'pip install -r source/lambda/testing_requirements.txt'
-pip install -r source/lambda/testing_requirements.txt
-echo 'cd source/lambda && pytest tests && cd ../deployment'
-cd source/lambda && pytest ./tests && cd ../../deployment
\ No newline at end of file
+#
```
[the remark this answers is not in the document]
CrypticCabub: Added a story to go through all files and fix all the headers together. I'm certain these are not the only ones

deployment/run-unit-tests.sh
```diff
@@ -1,6 +1,58 @@
...
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  or in the 'license' file accompanying this file. This file is distributed on an 'AS IS' BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions
+#  and limitations under the License.
+#
+
+
+set -e #fail script if any commands fail
```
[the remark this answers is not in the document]
CrypticCabub: valid option, but I'll probably be killing this script entirely in favor of running the tox commands directly in the buildspec.yml once I update the build process

deployment/run-unit-tests.sh
```diff
@@ -1,6 +1,58 @@
 #!/bin/bash
```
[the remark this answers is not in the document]
CrypticCabub: fair enough, I'll include it in the minor patch pr that will immediately follow this one

hearde (approved):

PR #376, opened 2023-02-15
Make state returns more explicit

CrypticCabub:
A tiny refactor to make the returned type more explicit -- ping me on slack for the diagram breaking this down if needed


CrypticCabub:
needs better testing to be in place before any changes to such core logic is safe to do

PR #380, opened 2023-02-27
setup github actions to run unit tests

CrypticCabub:
Add a github workflow that automatically runs unit tests on push/pull-request

The workflow will test the lambda python code using version 3.9, 3.10, and 3.11

This is adapted from the Refreezer project, and the unit tests all appear to run correctly. However, I am not sure if caching .mypy_cache is correct for this project (line 33 of the workflow). 

### DIFF
--- a/.github/workflows/unit-test-workflow.yml
+++ b/.github/workflows/unit-test-workflow.yml
@@ -1,0 +1,36 @@
+name: Unit test workflow
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        python: ["3.9", "3.10", "3.11"]
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: ${{ matrix.os }} ${{ matrix.python }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python }}
+          cache: 'pip'
+          cache-dependency-path: 'source/app/pyproject.toml'
+      - name: Install Tox and any other packages
+        run: |
+          pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


.github/workflows/unit-test-workflow.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        python: ["3.9", "3.10", "3.11"]
```
hearde, line 13: What's the reason to run the matrix on all three of these? Just to make sure you're writing forward-compatible code? The lambda in theory should run on exactly one version, the CLI I suppose should be tested against all currently supported versions.

.github/workflows/unit-test-workflow.yml
```diff
@@ -0,0 +1,36 @@
...
+          pip install tox-gh-actions
+      - name: Cache tox and mypy
+        uses: actions/cache@v3
+        with:
+          path: |
+            .tox
+            .mypy_cache
+          key: ${{ matrix.os }}-${{ matrix.python}}-tox-${{ hashFiles('pyproject.toml') }}-${{ hashFiles('tox.ini') }}-${{ hashFiles('.github/workflows/unit-test-workflow.yml') }}
+      - name: Run Tox
+        run: tox
```
hearde, line 36: Please configure your editor to terminate files with a newline

hearde (approved):

.github/workflows/unit-test-workflow.yml
CrypticCabub: this was how refreezer was setup and I saw no reason to remove it. Generally yes, I think the idea is to just make sure there are no weird breakages between versions to avoid being caught out in the long-term

PR #383, opened 2023-03-01
Add synth workflow for cdk nag (step 1)

CrypticCabub:
Add a workflow to automatically run cdk synth on push/PR

This is the first step to running CDK nag automatically on all pushes/PRs. The next step is to implement cdk nag into the templates themselves which will be the focus of the next PR

### DIFF
--- a/.github/workflows/cdk-nag-workflow.yml
+++ b/.github/workflows/cdk-nag-workflow.yml
@@ -1,0 +1,22 @@
+# Workflow that runs unit test
+name: CDK Nag
+
+on: [push, pull_request]
+
+jobs:
+  cdk-nag:
+    name: CDK Nag Check
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [16.x]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: CDK Synth
+        working-directory: source/infrastructure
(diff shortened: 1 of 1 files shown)


.github/workflows/cdk-nag-workflow.yml
```diff
@@ -0,0 +1,20 @@
...
+    strategy:
+      matrix:
+        node-version: [16.x]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+      - run: |
+          cd source/infrastructure && npm i
```
hearde, line 19: To avoid having to `cd`, you can specify:

```
defaults:
      run:
        working-directory: source
```

Also, use [npm ci](https://docs.npmjs.com/cli/v9/commands/npm-ci) for the following reasons:
- it will fail if package-lock is missing
- it will fail if package-lock and package.json are not reconcile-able - `npm i` may resolve differences by modifying package or package-lock, so package-lock does not reflect what was actually used in the workflow

hearde (changes requested):

.github/workflows/cdk-nag-workflow.yml
CrypticCabub: good catch. Will have to update the release guidance doc with this as well as I pretty much copied-pasted exactly what was in them for this

hearde (approved):

tmekari (approved):

PR #385, opened 2023-03-02
move tsc output to /deployment/cdk.dist

CrypticCabub:
I thought this was going to be harder...

Moving typescript build output to separate folder to prevent hidden build files overriding cdk commands run from cli during development

also updated the target ES version to 2022 since I was in the file anyway.

Tested by running build-s3-dist.sh locally to confirm that all output appear normal.

### DIFF
--- a/deployment/build-s3-dist.sh
+++ b/deployment/build-s3-dist.sh
@@ -78,10 +78,10 @@ echo "npm install"
 npm install
 
 # Run 'cdk synth' to generate raw solution outputs
-echo "cd "$cdk_source_dir""
+echo "cd $cdk_source_dir"
 cd "$cdk_source_dir"
 echo "node_modules/aws-cdk/bin/cdk synth --output=$staging_dist_dir"
-npm run build && node_modules/aws-cdk/bin/cdk synth --output=$staging_dist_dir --no-version-reporting
+node_modules/aws-cdk/bin/cdk synth --output=$staging_dist_dir --no-version-reporting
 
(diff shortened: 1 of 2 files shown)


G-Lenz (approved):

deployment/build-s3-dist.sh
```diff
@@ -78,10 +78,10 @@ echo "npm install"
 npm install
 
 # Run 'cdk synth' to generate raw solution outputs
-echo "cd "$cdk_source_dir""
+echo "cd $cdk_source_dir"
 cd "$cdk_source_dir"
 echo "node_modules/aws-cdk/bin/cdk synth --output=$staging_dist_dir"
-npm run build && node_modules/aws-cdk/bin/cdk synth --output=$staging_dist_dir --no-version-reporting
+node_modules/aws-cdk/bin/cdk synth --output=$staging_dist_dir --no-version-reporting
```
hearde, line 84: no change needed right now: why not npx? also I'd prefer that output and version-reporting be set in cdk.json rather than here. Actually it looks like output is already set there.

CrypticCabub: For npm vs npx I'm actually not too familiar with the difference between them and when exactly you would want to use one over another.

For the output directory, I think it's arguably better to define the output explicitly in this script rather than relying on the default (even if they are the same) because the rest of the script depends upon that output location. This way if the default ever changes for some reason, the build script does not break and vice-versa

PR #386, opened 2023-03-02
Feature/cdk nag

CrypticCabub:
Implementation of CDK nag and required suppressions

all suppressions should be reviewed and discussed before merging to see if any of them need to be added to the backlog for fiixing

### DIFF
--- a/source/infrastructure/lib/aws-instance-scheduler-stack.ts
+++ b/source/infrastructure/lib/aws-instance-scheduler-stack.ts
@@ -30,6 +30,7 @@ import { Construct } from "constructs";
 import { LambdaFunction } from "aws-cdk-lib/aws-events-targets";
 import {SUPPORTED_TIME_ZONES} from "./time-zones";
 import { AppRegistryForInstanceScheduler } from './app-registry';
+import {NagSuppressions} from "cdk-nag";
 
 
 export interface AwsInstanceSchedulerStackProps extends cdk.StackProps {
@@ -244,15 +245,14 @@ export class AwsInstanceSchedulerStack extends cdk.Stack {
       }
     }
 
-    //Start instance scheduler scheduler role reference and related references of principle, policy statement, and policy document.
+    //Start scheduler role reference and related references of principle, policy statement, and policy document.
     const compositePrincipal = new iam.CompositePrincipal(new iam.ServicePrincipal('events.amazonaws.com'), new iam.ServicePrincipal('lambda.amazonaws.com'))
 
     const schedulerRole = new iam.Role(this, "SchedulerRole", {
       assumedBy: compositePrincipal,
       path: '/'
     })
-
-    //End instance scheduler scheduler role reference
+    //End scheduler role reference
 
     //Start instance scheduler encryption key and encryption key alias.
     const instanceSchedulerEncryptionKey = new kms.Key(this, "InstanceSchedulerEncryptionKey", {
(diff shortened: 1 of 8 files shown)


source/infrastructure/lib/aws-instance-scheduler-stack.ts
```diff
@@ -708,6 +726,11 @@ export class AwsInstanceSchedulerStack extends cdk.Stack {
       }
     }
 
+    NagSuppressions.addResourceSuppressions(ec2DynamoDBPolicy_cfn_ref, [{
+      id: "AwsSolutions-IAM5",
+      reason: "This Lambda function has permission provided to write to CloudWatch logs using the iam roles."
```
hearde, line 731: The policy in question doesn't only have wildcard permissions to write logs, it's to get ssm parameters, describe ec2 and rds instances, get ssm maintenance windows, and get tags.

It's actually easier to review these from the snapshots.

source/infrastructure/lib/aws-instance-scheduler-stack.ts
```diff
@@ -574,6 +574,11 @@ export class AwsInstanceSchedulerStack extends cdk.Stack {
       roles: [schedulerRole]
     })
 
+    NagSuppressions.addResourceSuppressions(ec2Permissions, [{
+      id: "AwsSolutions-IAM5",
+      reason: "This Lambda function has permission provided to write to CloudWatch logs using the iam roles."
```
hearde, line 579: This policy is for modifying instances.

source/infrastructure/lib/aws-instance-scheduler-stack.ts
```diff
@@ -686,6 +700,10 @@ export class AwsInstanceSchedulerStack extends cdk.Stack {
         ]
       }
     }
+    NagSuppressions.addResourceSuppressions(lambdaToDynamoDb, [{
+      id: "AwsSolutions-IAM5",
+      reason: "This Lambda function has permission provided to write to CloudWatch logs using the iam roles."
+    }], true)
```
hearde, line 706: Not a big fan of applying to every child resource. This gets added to some strange things like the state table. You should probably go into the construct and add the required suppressions on the specific resources. When adding to a policy, review what the policy is for and explain. Someone is going to look at the suppression reason in a csv file, so it should generally explain why the permission is needed.

hearde (changes requested):

source/infrastructure/lib/aws-instance-scheduler-stack.ts
CrypticCabub: Actually It looks like I copy-pasted the wrong thing here. updated to match what the cfn-nag says

source/infrastructure/pipeline/pipeline-stack.ts
```diff
@@ -51,6 +52,28 @@ class PipelineStack extends Stack {
...
+    NagSuppressions.addStackSuppressions(this, [
+      {
+        id: "AwsSolutions-IAM5",
+        reason: "necessary permissions for being able to self-mutate"
+      },
+      {
+        id: "AwsSolutions-CB4",
+        reason: "Update step provided by construct"
+      }
+    ],true)
```
hearde, line 69: I would rather add these to the resources that require them than add to all children, even if it's in the pipeline. Are there a lot of findings that need to be suppressed? Or are they difficult to suppress because they change when the project changes? If that's the case we should probably add these as stack-level suppressions.

Also please configure your editor to insert semicolons.

CrypticCabub: I'll take a look and try to remove the child flags. I think I just copy-pasted it to all suppressions because it was the only way I found to get one of the earlier suppressions to work and didn't really think beyond it (have fixed that first suppression now)

hearde (approved):

PR #387, opened 2023-03-02
update cdk test to use ci and skip the unnecessary build step

CrypticCabub:
the same change was made to the build script in an earlier PR, this just brings the tox test script in-line and makes running local tests that much faster.


tox.ini
```diff
@@ -30,6 +30,5 @@ change_dir = source/infrastructure
 description = run CDK unit tests
 package = skip
 commands =
-    npm install
-    npm run build
+    npm ci
```
hearde, line 33: I'm not totally sure you want `ci` in here.`ci` is a clean install, which is good for github workflows and pipelines, but it's actually slower than `install` because of that. In ASR, `npm install` after already installing takes less than a second, while `ci` takes ~20 seconds. You're looking at a longer local test cycle. I'm not sure what the right way to decouple the local workflow from the github workflow is. Maybe ask Nathaniel.

CrypticCabub: yeah fair enough, I'll leave it as npm install and cut the run build.

difference between using ci and install appears to be ~30-40 and ~20 seconds respectively. I wonder if there's anything else we can do to bring that time down...

hearde (commented):
One impact of doing this is that since test is run with ts-jest, anything that's not tested is not built. So you could potentially push a breaking code change that's not caught by any workflow, if it's not run by the tests. If you're going to remove build from this ini, I would consider adding a workflow to just run `npm run build`. You could add it as a parallel job in the unit test workflow. Here's an example from ASR: https://github.com/aws-solutions/aws-security-hub-automated-response-and-remediation/blob/develop/.github/workflows/build.yml

hearde (approved):

PR #390, opened 2023-03-03
Update workflow names

CrypticCabub:
Update the names of github workflows to make it clearer what they are doing

### DIFF
--- a/.github/workflows/pull-request-workflow.yml
+++ b/.github/workflows/pull-request-workflow.yml
@@ -1,4 +1,4 @@
-name: Pull Request Workflow
+name: Viperlight
 
 on: 
   pull_request:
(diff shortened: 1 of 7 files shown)


.github/workflows/pull-request-workflow.yml
```diff
@@ -1,4 +1,4 @@
-name: Pull Request Workflow
+name: Viperlight
 
 on: 
   pull_request:
```
CrypticCabub, line 4: This currently only runs on pull_requests. Should it be updated to run on pushes as well?

hearde: I don't see a reason not to

hearde (approved):

.github/workflows/pull-request-workflow.yml
CrypticCabub: k. updating

tmekari (approved):

PR #391, opened 2023-03-06
Extract Core Scheduler into its own helper file

CrypticCabub:
This refactoring isolates the core scheduling lambda and related dynamo tables to simplify updating lambda deployment for the upcoming CDK-helper integration

--Snapshot tests pass without any changes to the snapshot--

### DIFF
--- a/source/infrastructure/lib/aws-instance-scheduler-stack.ts
+++ b/source/infrastructure/lib/aws-instance-scheduler-stack.ts
@@ -15,22 +15,22 @@
  *****************************************************************************/
 
 import * as cdk from 'aws-cdk-lib';
+import {Aws, RemovalPolicy} from 'aws-cdk-lib';
 import * as kms from 'aws-cdk-lib/aws-kms';
 import * as iam from 'aws-cdk-lib/aws-iam';
+import {ArnPrincipal, Effect, PolicyStatement} from 'aws-cdk-lib/aws-iam';
 import * as logs from 'aws-cdk-lib/aws-logs';
 import * as lambda from 'aws-cdk-lib/aws-lambda';
 import * as s3 from 'aws-cdk-lib/aws-s3';
 import * as sns from 'aws-cdk-lib/aws-sns';
 import * as events from 'aws-cdk-lib/aws-events';
-import { ArnPrincipal, Effect, PolicyStatement } from 'aws-cdk-lib/aws-iam';
-import { LambdaToDynamoDBProps, LambdaToDynamoDB } from '@aws-solutions-constructs/aws-lambda-dynamodb';
+import {Construct} from "constructs";
+import {LambdaFunction} from "aws-cdk-lib/aws-events-targets";
 import * as dynamodb from 'aws-cdk-lib/aws-dynamodb';
-import { Aws, RemovalPolicy } from 'aws-cdk-lib';
-import { Construct } from "constructs";
-import { LambdaFunction } from "aws-cdk-lib/aws-events-targets";
 import {SUPPORTED_TIME_ZONES} from "./time-zones";
 import { AppRegistryForInstanceScheduler } from './app-registry';
 import {NagSuppressions} from "cdk-nag";
+import {CoreScheduler} from "./core-scheduler";
 
 
 export interface AwsInstanceSchedulerStackProps extends cdk.StackProps {
@@ -48,8 +48,8 @@ export interface AwsInstanceSchedulerStackProps extends cdk.StackProps {
 /*
 * AWS instance scheduler stack, utilizes two cdk constructs, aws-lambda-dynamodb and aws-events-rule-lambda.
 * The stack has three dynamoDB tables defined for storing the state, configuration and maintenance information.
-* The stack also includes one lambda, which is scheduled using a AWS CloudWatch Event Rule. 
-* The stack also includes a cloudwatch log group for the entire solution, encrycption key, encyrption key alias and SNS topic,
+* The stack also includes one lambda, which is scheduled using an AWS CloudWatch Event Rule.
+* The stack also includes a cloudwatch log group for the entire solution, encryption key, encryption key alias and SNS topic,
 * and the necessary AWS IAM Policies and IAM Roles. For more information on the architecture, refer to the documentation at
 * https://aws.amazon.com/solutions/implementations/instance-scheduler/?did=sl_card&trk=sl_card
 */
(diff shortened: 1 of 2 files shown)


source/infrastructure/lib/aws-instance-scheduler-stack.ts
```diff
@@ -15,22 +15,22 @@
  *****************************************************************************/
 
 import * as cdk from 'aws-cdk-lib';
+import {Aws, RemovalPolicy} from 'aws-cdk-lib';
```
hearde, line 18: Can we be more consistent with the spacing of the destructuring syntax? Follow the pattern established in the file.

source/infrastructure/lib/aws-instance-scheduler-stack.ts
```diff
@@ -15,22 +15,22 @@
  *****************************************************************************/
```
hearde, line 15: Please replace the long copyright headers with the short ones as you encounter them.

source/infrastructure/lib/aws-instance-scheduler-stack.ts
```diff
@@ -15,22 +15,22 @@
...
 
 import * as cdk from 'aws-cdk-lib';
+import {Aws, RemovalPolicy} from 'aws-cdk-lib';
 import * as kms from 'aws-cdk-lib/aws-kms';
 import * as iam from 'aws-cdk-lib/aws-iam';
+import {ArnPrincipal, Effect, PolicyStatement} from 'aws-cdk-lib/aws-iam';
```
hearde, line 21: Please either convert all `aws-iam` imports to named imports or use the namespace imports from `iam`. Multiple instances.

source/infrastructure/lib/aws-instance-scheduler-stack.ts
```diff
@@ -296,141 +296,33 @@ export class AwsInstanceSchedulerStack extends cdk.Stack {
...
-        'dynamodb:BatchWriteItem'
-      ],
-      effect: Effect.ALLOW,
-      resources: [
-        cfnConfigTable.attrArn,
-        cfnMaintenanceWindowTable.attrArn
-      ]
+    //instance scheduler core scheduler construct reference.
+    const coreScheduler = new CoreScheduler(this,  {
+      solutionVersion: props["solutionVersion"],
```
hearde, line 301: Use dot notation unless you need bracket notation.

source/infrastructure/lib/core-scheduler.ts
```diff
@@ -0,0 +1,159 @@
+import {LambdaToDynamoDB, LambdaToDynamoDBProps} from "@aws-solutions-constructs/aws-lambda-dynamodb";
```
hearde, line 1: Missing copyright header.

source/infrastructure/lib/core-scheduler.ts
```diff
@@ -0,0 +1,159 @@
...
+import {Aws, RemovalPolicy} from "aws-cdk-lib";
+import * as lambda from "aws-cdk-lib/aws-lambda";
+import * as cdk from "aws-cdk-lib";
+import * as dynamodb from "aws-cdk-lib/aws-dynamodb";
+import {Construct} from "constructs";
+import {Effect, PolicyStatement} from "aws-cdk-lib/aws-iam";
+import {NagSuppressions} from "cdk-nag";
+export interface InstanceSchedulerLambdaProps {
```
hearde, line 9: Formatting. Put a newline above this and get rid of the one below it.

source/infrastructure/lib/core-scheduler.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  solutionVersion: string
```
hearde, line 11: readonly - this was commented in the previous PR - please address all of those comments before re-requesting review: https://github.com/aws-solutions/aws-instance-scheduler/pull/374

hearde (changes requested):

source/infrastructure/lib/aws-instance-scheduler-stack.ts
CrypticCabub: I would rather do this as a separate story/pr and change them all at once using an IDE macro

source/infrastructure/lib/core-scheduler.ts
```diff
@@ -0,0 +1,159 @@
+import {LambdaToDynamoDB, LambdaToDynamoDBProps} from "@aws-solutions-constructs/aws-lambda-dynamodb";
+import {Aws, RemovalPolicy} from "aws-cdk-lib";
+import * as lambda from "aws-cdk-lib/aws-lambda";
```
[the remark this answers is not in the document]
CrypticCabub: as I understand it namespace imports are the "cdk way" (at least that's how the cdk bootcamp does things) so I switched as much as I could reasonably move over to namespace imports and left the ones that didn't make sense to port

source/infrastructure/lib/core-scheduler.ts
```diff
@@ -0,0 +1,159 @@
...
+        billingMode: dynamodb.BillingMode.PAY_PER_REQUEST,
+        removalPolicy: RemovalPolicy.DESTROY,
+        pointInTimeRecovery: true
+      },
+      tablePermissions: "ReadWrite",
+    }
+
+  }
+
+  // const dependenciesLayer = new lambda.LayerVersion() {
```
hearde, line 155: remove

source/infrastructure/lib/core-scheduler.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    const lambdaToDynamoDbConstruct = new LambdaToDynamoDB(scope, 'instance-scheduler-lambda', this.lambdaToDynamoPropsFrom(props))
+
+    this.lambdaFunction = lambdaToDynamoDbConstruct.lambdaFunction;
+    this.stateTable = lambdaToDynamoDbConstruct.dynamoTable;
+
+    const cfnStateTable = this.stateTable.node.defaultChild as dynamodb.CfnTable
+    cfnStateTable.overrideLogicalId('StateTable')
+    cfnStateTable.addPropertyOverride("SSESpecification", {
+      "KMSMasterKeyId": { "Ref": "InstanceSchedulerEncryptionKey" },
```
hearde, line 42: Please pass in the key and use `Fn.ref` on the object

source/infrastructure/lib/core-scheduler.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    this.lambdaFunction.addToRolePolicy(dynamodbPolicy)
+
+    NagSuppressions.addResourceSuppressions(lambdaToDynamoDbConstruct.node.findChild("LambdaFunctionServiceRole"), [{
+      id: "AwsSolutions-IAM5",
+      reason: "This Lambda function needs to be able to write a log streams for each scheduler execution (1 per account/region/service)"
+    }])
+  }
+
+  private lambdaToDynamoPropsFrom(props: InstanceSchedulerLambdaProps) : LambdaToDynamoDBProps {
```
hearde, line 120: I don't think this needs to be a member function. Also, the name confuses me, but I'm not sure what would be better because of the overloaded terms "construct" and "props".

source/infrastructure/lib/core-scheduler.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    //to be able to cdk deploy:
+    //lambda.Code.fromAsset("../app")
```
hearde, line 123: What does this mean?

source/infrastructure/lib/core-scheduler.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    return {
+     lambdaFunctionProps: {
+       functionName: Aws.STACK_NAME + '-InstanceSchedulerMain',
+       description: 'EC2 and RDS instance scheduler, version ' + props["solutionVersion"],
```
hearde, line 128: Dot notation

source/infrastructure/lib/core-scheduler.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    const cfnStateTable = this.stateTable.node.defaultChild as dynamodb.CfnTable
+    cfnStateTable.overrideLogicalId('StateTable')
+    cfnStateTable.addPropertyOverride("SSESpecification", {
+      "KMSMasterKeyId": { "Ref": "InstanceSchedulerEncryptionKey" },
+      "SSEEnabled": true,
+      "SSEType": 'KMS'
+    })
+
+    this.configTable = new dynamodb.Table(scope, 'ConfigTable', {
```
hearde, line 47: The table properties are repeated three times in this file. Can we consolidate them?

hearde (commented):
I would rather you add fine-grained tests for this class. I don't consider snapshot tests alone to be sufficient coverage.

source/infrastructure/lib/core-scheduler.ts
CrypticCabub: if memory serves, it was a note to say to confirm which call would be needed to support cdk deploy to save time on the followup update that was planned

source/infrastructure/lib/core-scheduler.ts
CrypticCabub: the tables have subtly different properties and configurations. Perhaps they could be better consolidated in a followup refactor? (perhaps also removing the LambdaToDynamoDB construct like we had discussed this a.m.)

source/infrastructure/lib/core-scheduler.ts
CrypticCabub: renamed to extractLambdaToDynamoPropsFrom(). 

As for the pros/cons of having the function at all. I feel it allows the code to more clearly document exactly what is happening. I can look at the high-level resources being created and not get distracted by a huge parameter list that would otherwise distract from what's actually going on

source/infrastructure/lib/aws-instance-scheduler-stack.ts
```diff
@@ -48,7 +48,7 @@ export interface AwsInstanceSchedulerStackProps extends cdk.StackProps {
 /*
 * AWS instance scheduler stack, utilizes two cdk constructs, aws-lambda-dynamodb and aws-events-rule-lambda.
 * The stack has three dynamoDB tables defined for storing the state, configuration and maintenance information.
-* The stack also includes one lambda, which is scheduled using a AWS CloudWatch Event Rule. 
+* The stack also includes one lambda, which is scheduled using an AWS CloudWatch Event Rule.
 * The stack also includes a cloudwatch log group for the entire solution, encrycption key, encyrption key alias and SNS topic,
```
tmekari, line 52: Wouldn't have noticed this without the change above, but encryption is spelled wrong here a couple of times. Not something to fix now but just something I noticed.

tmekari (approved):

PR #395, opened 2023-03-06
Add tests for core scheduler

hearde:
- add fine-grained cdk tests for core scheduler

Ideological question: do these tests add value? Or are they a waste of time?

### DIFF
--- a/source/infrastructure/jest.config.ts
+++ b/source/infrastructure/jest.config.ts
@@ -1,16 +1,17 @@
 module.exports = {
-  "roots": [
-    "<rootDir>/tests"
-  ],
-  testMatch: [ '**/*.test.ts'],
+  roots: ["<rootDir>/tests", "<rootDir>/lib"],
+  testMatch: ["**/*.test.ts"],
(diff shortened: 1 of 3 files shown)


CrypticCabub:
Thanks for taking the time to put this together.

From our discussion before, I don't think it's fair to say that they don't add any value as you are right that they make it much clearer what has actually changed when snapshot diffs are large, But at the same time tests like this are essentially just testing the CDK -> CFN transformation and appear to roughly triple the amount of code for any given component.

The way I see it, the pros could be argued to be:
- somewhat faster identification of exactly intended behavior changed between two snapshots
- mechanism for forcing one to acknowledge every behavioral change in the template rather than just running a snapshot update
- forcing one to think about exactly why a certain decision has been made?

The question would be whether or not we already have these things by comparing diffs of the CDK code itself in conjunction with the snapshot diff, and if the substantial increase in number of code lines to maintain is worth any gains we might get in this area

CrypticCabub (approved):

PR #401, opened 2023-03-09
Feature/pipeline deploy

CrypticCabub:
This grew a little out of hand thanks to interconnected build dependencies and a few mistakes on my end with the git history and merge timings

**CDK-Helper Deployment:**

The solution now deploys via native cdk-deploy and the build-s3-dist script leverages CDK-Helper to grab the cdk lambda asset for s3-bunding in our internal pipelines. This involved several structural refactorings and improvements (detailed below)

**@aws-cdk/aws-lambda-python-alpha module**
Packaging of the lambda code was originally done manually by the build script, but this is not possible when supporting cdk-deploy. Instead, packaging is now done with the L2 python lambda construct which leverages a docker container to package the lambda dependencies during the cdk-synth step.

This module does not support pyproject.toml natively and so bundling is currently achieved using the "beforeBundling" hook of the construct. This may make the L2 construct redundant and replaceable with the L1 Lambda equivalent by just adding a custom bundling step, but I have not yet had the time to delve into this option deeper.

**cdk-context abstraction**
Static solution variables have been extracted into cdk.json context variables and are now retrieved through the cdk-context access abstraction. Because cdk-helper deployment involves detecting pipeline environment variables to correctly setup the lambda source location, this changes allows the static solution variables to be kept separate and not interfere with the environment variables used by cdk-helper

**rewritten build script**
The build script was rewritten to support CDK-helper, and respect the dependencies (including cdk) defined in package.json.

Previously, the build-script included a cdk-version that it would overwrite the package.json with via an npm install. this has been removed entirely

minor issue: the original build-script would automatically update version.py in the lambda which would provide the current solution version to certain logging strings. With the new deployment, this is no longer possible and version.py has been updated to hard-code the current solution version. A story should probably be created to redesign this version.py file into possibly an environment variable which can be updated automatically by cdk

**Unification of dependencies into pyproject.toml**
dependencies were split across requirements.txt and pyproject.toml. With the rewriting of the build-script, I took the opportunity to re-unify them into a single source of truth


deployment/build-s3-dist.sh
```diff
@@ -22,20 +24,26 @@ if [ -z "$1" ] || [ -z "$2" ] || [ -z "$3" ]; then
...
 # Get reference for all important folders
-template_dir="$PWD"
-staging_dist_dir="$template_dir/staging"
-global_dist_dir="$template_dir/global-s3-assets"
-regional_dist_dir="$template_dir/regional-s3-assets"
-lambda_source_dir="$template_dir/../source/app"
-cli_source_dir="$template_dir/../source/cli"
-cdk_source_dir="$template_dir/../source/infrastructure"
+project_root="$PWD/.."
+deployment_dir="$PWD"
```
hearde, line 33: No need to change this now, but in general I prefer to base paths relative to the file rather than the calling location. This can be done with something like: `local root_dir=$(dirname "$(cd -P -- "$(dirname "$0")" && pwd -P)")`

hearde (commented):
Haven't reviewed very much, will get back tothis

deployment/build-s3-dist.sh
```diff
@@ -46,58 +54,32 @@ echo "--------------------------------------------------------------------------
 echo "[Init] Remove any old dist files from previous runs"
```
hearde, line 54: duplicate set -x
recommend set -o pipefail as well

source/infrastructure/bin/aws-instance-scheduler.ts
```diff
@@ -18,48 +18,47 @@ import * as cdk from 'aws-cdk-lib';
...
-const SOLUTION_ID = process.env['SOLUTION_ID'] ? process.env['SOLUTION_ID'] : "SO0030";
-const SOLUTION_BUCKET = process.env['DIST_OUTPUT_BUCKET'] ? process.env['DIST_OUTPUT_BUCKET'] : "";
-const SOLUTION_TMN = process.env['SOLUTION_TRADEMARKEDNAME'] ? process.env['SOLUTION_TRADEMARKEDNAME'] : "aws-instance-scheduler";
-const SOLUTION_PROVIDER = 'AWS Solution Development';
-const APP_REG_APPLICATION_TYPE = process.env['APP_REG_APPLICATION_TYPE'] ? process.env['APP_REG_APPLICATION_TYPE'] : 'AWS-Solutions';
-const APP_REG_SOLUTION_NAME = process.env['APP_REG_SOLUTION_NAME'] ? process.env['APP_REG_SOLUTION_NAME'] : "instance-scheduler-on-aws";
+
+let synthesizer = new DefaultStackSynthesizer({
+    generateBootstrapVersionRule: false,
+});
```
hearde, line 28: Why not let the stack that's intended to be cdk-deployed generate the bootstrap rule?

source/infrastructure/bin/aws-instance-scheduler.ts
```diff
@@ -18,48 +18,47 @@ import * as cdk from 'aws-cdk-lib';
...
+
+// Solutions pipeline deployment
+const { DIST_OUTPUT_BUCKET, SOLUTION_NAME, VERSION } = process.env;
+if (DIST_OUTPUT_BUCKET && SOLUTION_NAME && VERSION)
```
hearde, line 32: use curly braces

source/infrastructure/bin/aws-instance-scheduler.ts
```diff
@@ -18,48 +18,47 @@ import * as cdk from 'aws-cdk-lib';
...
+    synthesizer = new DefaultStackSynthesizer({
+        generateBootstrapVersionRule: false,
+        fileAssetsBucketName: `${DIST_OUTPUT_BUCKET}-\${AWS::Region}`,
```
hearde, line 35: is this backslash needed? I can't understand why

source/infrastructure/lib/core-scheduler.ts
```diff
@@ -128,42 +168,42 @@ export class CoreScheduler {
...
     this.lambdaFunction.addToRolePolicy(dynamodbPolicy)
 
-    NagSuppressions.addResourceSuppressions(lambdaToDynamoDbConstruct.node.findChild("LambdaFunctionServiceRole"), [{
-      id: "AwsSolutions-IAM5",
-      reason: "This Lambda function needs to be able to write a log streams for each scheduler execution (1 per account/region/service)"
-    }])
+    // NagSuppressions.addResourceSuppressions(lambdaToDynamoDbConstruct.node.findChild("LambdaFunctionServiceRole"), [{
+    //   id: "AwsSolutions-IAM5",
+    //   reason: "This Lambda function needs to be able to write a log streams for each scheduler execution (1 per account/region/service)"
+    // }])
```
hearde, line 174: remove commented out code

source/infrastructure/package.json
```diff
@@ -23,6 +23,7 @@
   },
   "dependencies": {
     "@aws-cdk/aws-servicecatalogappregistry-alpha": "^2.62.2-alpha.0",
+    "@aws-cdk/aws-lambda-python-alpha" : "~2.68.0-alpha.0",
     "@aws-solutions-constructs/aws-lambda-dynamodb": "2.30.0",
```
hearde, line 27: match the versions of all the cdk deps

hearde (changes requested):

deployment/build-s3-dist.sh
```diff
@@ -12,8 +12,10 @@
 #  and limitations under the License.
 #
 
-# Important: CDK global version number
-cdk_version=2.62.2
+
+#"$DEBUG" == 'true' -> enable detailed messages.
+#set -x enables a mode of the shell where all executed commands are printed to the terminal
+[ "$DEBUG" == 'true' ] && set -x
```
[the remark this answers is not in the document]
CrypticCabub: Thanks, updating...

deployment/build-s3-dist.sh
```diff
@@ -22,20 +24,26 @@ if [ -z "$1" ] || [ -z "$2" ] || [ -z "$3" ]; then
     exit 1
 fi
 
-export DIST_VERSION=$3
 export DIST_OUTPUT_BUCKET=$1
-export SOLUTION_ID=SO0030
 export SOLUTION_NAME=$2
-export SOLUTION_TRADEMARKEDNAME=$2
+export VERSION=$3
```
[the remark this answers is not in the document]
CrypticCabub: Good catch. the DIST_VERSION usage was for the CLI which I left as-is and never manually tested, we should add cli tests as a story item to at least check before release to ensure none of the other changes over the last couple months have impacted it

"VERSON" was because that was the value expected in the helper code added to aws-instancescheduler.ts I have updated both to use "DIST_VERSION" to be more explicit, though I wonder if this is the correct name or not. The purpose of the variable is to refer to the version argument passed in by the internal pipeline and is (supposed to be) only used to produce the object name in the solutions deployment bucket. However, it looks like the CLI also takes this version in for some unknown purpose inside of setuptools (might just be to set the metadata tag on the package)

source/infrastructure/bin/aws-instance-scheduler.ts
CrypticCabub: This was how it was done in SIH, however I think I agree with this approach as it allows the stack to be built as if it was a normal cdk stack without worrying about the mutations necessary to support the solutions pipeline

source/infrastructure/bin/aws-instance-scheduler.ts
CrypticCabub: what would be the specific reason for this?

source/infrastructure/bin/aws-instance-scheduler.ts
CrypticCabub: It's to escape the $ so that CFN can fill in the region. (or so I presume, I copy-pasted this from SIH)

source/infrastructure/bin/aws-instance-scheduler.ts
hearde: just a style thing in this case, but the reason is that if someone adds a second line in this conditional, it won't work like you expect. better to be explicit and not make it possible for someone to make a mistake in the future.

source/infrastructure/lib/core-scheduler.ts
```diff
@@ -48,9 +47,50 @@ export class CoreScheduler {
...
+      role: props.schedulerRole,
+      memorySize: props.memorySize,
+      timeout: cdk.Duration.seconds(300),
+      environment: props.environment,
+      tracing: lambda.Tracing.ACTIVE,
+
+      bundling: {
+        commandHooks: {
+          beforeBundling(inputDir: string, outputDir: string): string[] {
+             return [`pip install --target ${outputDir} -e ${inputDir}`]
```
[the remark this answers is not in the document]
CrypticCabub: hmm, this may be the origin of the "editable" files I had been seeing. I had assumed the -e was required for specifying the input dir but I may have been mistaken. I'll try removing it and see what it does

source/infrastructure/lib/core-scheduler.ts
CrypticCabub: Thank you!

source/infrastructure/package.json
CrypticCabub: lambda-dynamodb construct appears to follow a different version pattern, but I have updated aws-cdk-lib and @aws-cdk/aws-servicecatalogappregistry-alpha accordingly

source/infrastructure/tests/__snapshots__/aws-instance-scheduler-stack.test.ts.snap
```diff
@@ -1894,71 +1885,6 @@ exports[`AwsInstanceSchedulerStack snapshot test 1`] = `
       "Type": "AWS::DynamoDB::Table",
       "UpdateReplacePolicy": "Delete",
     },
-    "instanceschedulerlambdaLambdaFunctionServiceRoleEBF44CD1": {
```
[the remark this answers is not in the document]
CrypticCabub: evidently not. I didn't try to remove it explicitly so it was likely an artifact of allowing the lambda-dynamodb construct to initialize the lambda instead of providing a pre-made one

source/infrastructure/lib/core-scheduler.ts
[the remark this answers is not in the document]
CrypticCabub: yes, that's exactly what that was. good catch.

source/infrastructure/bin/aws-instance-scheduler.ts
```diff
@@ -18,48 +18,47 @@ import * as cdk from 'aws-cdk-lib';
...
-    solutionId: SOLUTION_ID,
-    solutionTradeMarkName: SOLUTION_TMN,
-    solutionProvider: SOLUTION_PROVIDER,
-    solutionBucket: SOLUTION_BUCKET,
-    solutionName: SOLUTION_NAME,
-    solutionVersion: SOLUTION_VERSION,
-    appregApplicationName: APP_REG_APPLICATION_TYPE,
-    appregSolutionName: APP_REG_SOLUTION_NAME
+    synthesizer: synthesizer,
+    description:  `(${solutionDetails.solutionId}) - The AWS CloudFormation template for ${solutionDetails.solutionName} cross account role, version: ${solutionDetails.solutionVersion}`,
```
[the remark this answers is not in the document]
CrypticCabub: oh yeah that, then why isn't it part of the ID in the first place?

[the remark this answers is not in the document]
hearde: It's the ID of the spoke stack

source/infrastructure/package.json
CrypticCabub: nevermind, updating these is non-trivial (particularly appregistry). I have created a separate story to address this

source/infrastructure/lib/core-scheduler.ts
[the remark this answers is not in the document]
CrypticCabub: Apparently losing this flag breaks the docker build of the lambda... Have re-added it and re-opened the story to investigate it more deeply

source/infrastructure/bin/aws-instance-scheduler.ts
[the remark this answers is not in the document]
CrypticCabub: fixed

deployment/build-s3-dist.sh
CrypticCabub: updated to include a full `set -eou pipefail` immediately after parsing input variables

hearde (approved):

tmekari (approved):

Issue #160, opened 2020-04-16
Support Elastic Container Service (Amazon ECS)


cbanciu667:
Hi,

Is there anything planned to handle ECS clusters and its ASGs ?
I worked on something similar in the past (see attachment). Maybe it can be used as inspiration.
[ecs-asg.py.txt](https://github.com/awslabs/aws-instance-scheduler/files/4484789/ecs-asg.py.txt)

gockle:
@cbanciu667 thank you for the input, this will be taken up as a new feature request.

takezoux2:
@maykays 
Why this feature is moved to "We won't build" section?
Are you developing this feature in another project? or is this feature beyond the responsibility of this project?

gockle:
Hi @takezoux2  The current design of instance scheduler solution has some limitations which needs to be addressed, we are currently evaluating a new/alternate design options for the solution, Hence at this point we don't want to include/expand this solution in its current architecture to more services. I would suggest you consider taking a look at [AWS Systems Manager Automation](https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-automation.html)

takezoux2:
Thanks😊 I check AWS System Manager Automation.

rostskadat:
Hello @gockle, @rakshb 

I understand your suggestion about the AWS Systems Manager Automation for this use case.
However the client I'm working is really interested in having the solution support AWS ECS Service, 
as  that would allow them a quick win in term of cost for their current environment.
I therefore submitted the Pull request #407 specifically for this issue.
Please, do no hesitate to comment anything that might allow a prompt merge of this request.

Thanks for the consideration

bientavu:
Hi @rakshb, we are also really interested for this new feature. Do you have an approximate ETA for this prioritized feature ?

CrypticCabub:
ASG Scheduling is now supported in [v3.0.0](https://github.com/aws-solutions/instance-scheduler-on-aws/releases/tag/v3.0.0), which by extension means that you can schedule the ASGs that run your ECS clusters.

villelahdenvuo:
@CrypticCabub Hey, I was looking into setting up org level instance scheduler and I would like to consolidate our ECS Fargate services to use the same schedule as other services, but it does not use ASG, seems like Fargate has its own scaling system?

Should I open a new issue for supporting Fargate services in the instance scheduler solution?

CrypticCabub:
[...]

PR #407, opened 2023-03-20
Support Elastic Container Service (Amazon ECS)

rostskadat:
Issue #160

*Description of changes:*
This pull request add support for the ECS Service. It works by setting the ECS Service's `desiredCount` to 0 when "stopping the instance" and reset it to its original value when "starting the instance". It saves the original `desiredCount `as a tag on the ECS Service. It allows our client to integrate the ECS Cluster in their environment and benefit from a quick win in term of cost saving.
Please note that it is helpful only in case of tasks of type `FARGATE`. In case the `launchType` is `EC2` or `EXTERNAL` the underlying compute resource is not scaled in.

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -23,7 +23,7 @@ phases:
   post_build:
     commands:
       - echo "Retrieving next stage buildspec `date` in `pwd`"
-      - aws s3 cp s3://solutions-build-assets/changelog-spec.yml ../buildspec.yml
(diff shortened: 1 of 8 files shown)


CrypticCabub:
Thanks @rostskadat! The team will need to review the overall design in light of our current roadmap but I have added an item to our current backlog to give it a proper review

PR #410, opened 2023-03-23
Feature/e2e

CrypticCabub:
Implement automated End to End Testing

This PR implements an automated end-to-end testing framework along with a basic first test that:
- deploys the solution
- deploys a test EC2 instance
- creates a basic test schedule
- confirms that the EC2 instance is started/stopped correctly in accordance with the schedule

All unit tests pass, and the end-to-end test also passes


source/infrastructure/pipeline/e2e-tests/basic-ec2-start-stop.test.ts
```diff
@@ -0,0 +1,65 @@
...
+import * as dynamodb from "@aws-sdk/client-dynamodb";
+
+import {resourceParams} from "./basic-ec2-start-stop.test.resources";
+import {delayMinutes} from "./index";
+import {getInstanceState} from "./utils/ec2-test-utils";
+import {createSchedule, currentTimePlus} from "./utils/schedule-test-utils";
+
+
+const ec2Client = new ec2.EC2Client({});
+const dynamoClient = new dynamodb.DynamoDBClient({});
```
hearde, line 25: Should we be using our solution custom user agent here? I'm a little torn because it's not really necessary in test code, but on the other hand if customers are running this pipeline we would want those metrics to show up.

source/infrastructure/pipeline/e2e-tests/utils/ec2-test-utils.ts
```diff
@@ -0,0 +1,28 @@
...
+    new ec2.DescribeInstanceStatusCommand({
+      InstanceIds: [instanceId],
+      IncludeAllInstances: true
+    })
+  )
+
+  if(!result.InstanceStatuses) throw new Error(`Instance with id of ${instanceId} Not Found`)
+
+  return result.InstanceStatuses![0].InstanceState!.Name
+}
```
hearde, line 28: Can we get a unit test for this?

source/infrastructure/pipeline/e2e-tests/utils/ec2-test-utils.ts
```diff
@@ -0,0 +1,28 @@
...
+  const result = await client.send(
+    new ec2.DescribeInstanceStatusCommand({
+      InstanceIds: [instanceId],
+      IncludeAllInstances: true
+    })
+  )
+
+  if(!result.InstanceStatuses) throw new Error(`Instance with id of ${instanceId} Not Found`)
+
+  return result.InstanceStatuses![0].InstanceState!.Name
```
hearde, line 27: Is assert not null correct here? If there's no response, we should probably throw an error. Though I guess that will happen anyway when you try to access a nonexistent property. You also already check, so I'm not sure why the assertion is necessary.

source/infrastructure/pipeline/e2e-tests/utils/schedule-test-utils.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+export async function createSchedule(client: dynamodb.DynamoDBClient, schedule: Schedule) {
+  await client.send(
+    new dynamodb.BatchWriteItemCommand({
+      RequestItems: {
+        [configTableName]: scheduleToBatchedPutRequests(schedule)
+      }
+    })
+  )
+}
```
hearde, line 39: Can we have some unit tests for these functions?

source/infrastructure/pipeline/lib/pipeline-stack.ts
```diff
@@ -1,18 +1,22 @@
...
 
-import CodeStarSource from "./code-star-source";
-import { Construct } from "constructs";
-import { CodeBuildStep, CodePipeline } from "aws-cdk-lib/pipelines";
-import { ComputeType, LinuxBuildImage } from "aws-cdk-lib/aws-codebuild";
-import { Stack, Stage } from "aws-cdk-lib";
-import { StringParameter } from "aws-cdk-lib/aws-ssm";
 import * as codebuild from "aws-cdk-lib/aws-codebuild";
-import { AwsInstanceSchedulerStack } from "../lib/aws-instance-scheduler-stack";
+import * as pipeslines from "aws-cdk-lib/pipelines"
```
hearde, line 5: spelling

source/infrastructure/pipeline/lib/pipeline-stack.ts
```diff
@@ -1,18 +1,22 @@
...
 import {NagSuppressions} from "cdk-nag";
+import {E2eTestStack} from "./e2e-test-stack";
+import {Effect, PolicyStatement} from "aws-cdk-lib/aws-iam";
+import * as hubStackUtils from "../e2e-tests/utils/hub-stack-utils";
 
-const DEPLOY_STAGE_NAME = "test-deploy";
+const DEPLOY_STAGE_NAME = "Deployment-Test";
+const END_TO_END_STAGE_NAME = "End-to-End-Tests"
 const STACK_NAME = "InstanceScheduler";
+const TEST_RESOURCES_STACK_NAME = "InstanceSchedulerE2ETestResources";
```
hearde, line 19: Any reason for these variables?

hearde (commented):
Really good working getting this all functional. If any of the suggestions here are too large in scope, I'd rather get this in and create follow-on tasks.

README.md
```diff
@@ -141,18 +145,18 @@ branch = develop
 ### Step 3 - Deploy the Testing Pipeline
 
 ```
-cd /source/infrastructure
+cd source/infrastructure
+npm install
+cd pipeline
 cdk bootstrap
-cdk deploy testing-pipeline
+cdk deploy aws-instance-scheduler-testing-pipeline
```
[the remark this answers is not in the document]
CrypticCabub: this particular naming is used in several places all over the solution. If it's relevant it should probably be done as a single sweeping PR that fixes it everywhere

source/infrastructure/instance-scheduler/bin/aws-instance-scheduler.ts
```diff
@@ -60,9 +59,6 @@ new AwsInstanceSchedulerRemoteStack(app, 'aws-instance-scheduler-remote', {
     appregSolutionName: solutionDetails.appRegSolutionName
 });
 
-new PipelineStack(app, 'aws-instance-scheduler-testing-pipeline');
```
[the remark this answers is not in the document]
CrypticCabub: isolating the solution itself from the "app" that is designed to test the solution. In particular this allows the solution to be built and published for deployment without having to also build the testing pipeline and related assets

source/infrastructure/instance-scheduler/lib/aws-instance-scheduler-stack.ts
```diff
@@ -39,6 +39,28 @@ export interface AwsInstanceSchedulerStackProps extends cdk.StackProps {
...
+  readonly trace?: "Yes"|"No",
+  readonly enableSSMMaintenanceWindows?: "Yes"|"No",
+  readonly tagName?: string,
+  readonly defaultTimezone?: string,
+  readonly regions?: string,
+  readonly crossAccountRoles?: string,
+  readonly startedTags?: string,
+  readonly stoppedTags?: string,
+  readonly schedulerFrequency?: "1"|"2"|"5"|"10"|"15"|"30"|"60",
+  readonly scheduleLambdaAccount?: "Yes"|"No",
```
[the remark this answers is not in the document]
CrypticCabub: I'm not sure I like the current handling of cfn parameters in general (duplication between the parameter and the interface). Perhaps they should get the same treatment as the core scheduler and be extracted into a separate component, but that would be out of the scope of this PR

source/infrastructure/instance-scheduler/lib/aws-instance-scheduler-stack.ts
```diff
@@ -51,6 +73,11 @@ export interface AwsInstanceSchedulerStackProps extends cdk.StackProps {
 */
 export class AwsInstanceSchedulerStack extends cdk.Stack {
 
+
+  readonly configurationTableOutput: CfnOutput;
+  readonly issueSnsTopicArn: CfnOutput;
+  readonly schedulerRoleArn: CfnOutput;
```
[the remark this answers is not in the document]
CrypticCabub: for the pipeline to be able to access it (and thus publish it to the testing environment), it must be able to access the CfnOutput object.

This particular approach is a quick-and-dirty way of doing that, but just like with the cfn parameters comment I would be tempted to extract all of the CfnOutputs into some sort of consolidated object to improve this in the future.

source/infrastructure/pipeline/bin/testing-pipeline.ts
```diff
@@ -0,0 +1,36 @@
...
+ *  not use this file except in compliance with the License. A copy of the    *
+ *  License is located at                                                     *
+ *                                                                            *
+ *      http://www.apache.org/licenses/LICENSE-2.0                            *
+ *                                                                            *
+ *  or in the 'license' file accompanying this file. This file is distributed *
+ *  on an 'AS IS' BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,        *
+ *  express or implied. See the License for the specific language governing   *
+ *  permissions and limitations under the License.                            *
+ *****************************************************************************/
```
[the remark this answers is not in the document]
CrypticCabub: planned for a separate PR. I just copy-pasted these from another file to make it easier to update everything at once later with a find-and-replace

source/infrastructure/instance-scheduler/lib/aws-instance-scheduler-stack.ts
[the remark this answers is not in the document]
hearde: But if you add a new parameter, you either don't allow it to be overridden through this interface, or you do by adding a new property. Neither of those scenarios is particularly problematic. What I'm concerned about is if the legal literals for a parameter (e.g. scheduler frequency) change, you have to change it in two places or you may end up with a deployment failure.

source/infrastructure/pipeline/bin/testing-pipeline.ts
```diff
@@ -0,0 +1,36 @@
...
+
+
+const app = new cdk.App();
+
+new PipelineStack(app, 'aws-instance-scheduler-testing-pipeline');
+
+/*
+E2eTestStack does not actually need to be built here to work in the pipeline,
+but building it here ensures it gets covered by CDK-Nag
+ */
```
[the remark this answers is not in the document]
CrypticCabub: That's a good question. I wrote this comment before we realized that the synth step was just secretly synthing everything anyway.

I suspect that this is still necessary though as the pipeline appears to synth stacks to an asset directory similar to the lambda construct

source/infrastructure/instance-scheduler/lib/aws-instance-scheduler-stack.ts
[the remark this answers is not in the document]
hearde: I'm saying that creating the `CfnOutput` explicitly may not be necessary. When you pass values across stacks, CDK creates outputs/parameters for you:

```ts
readonly configTableArn: string;
...
this.configTableArn = (coreScheduler.configTable.node.defaultChild as dynamodb.CfnTable).attrArn;
```

Again, not totally sure this will work, but it's worth trying.

source/infrastructure/pipeline/e2e-tests/basic-ec2-start-stop.test.resources.ts
```diff
@@ -0,0 +1,74 @@
...
+
+import {Construct} from "constructs";
+import {NagSuppressions} from "cdk-nag";
+import {TestResourceProvider} from "./index";
+
+
+
+const envKeys = {
+  ec2InstanceId: 'basic_start_stop_instance_id',
+}
```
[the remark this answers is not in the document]
CrypticCabub: I believe the type is inferred automatically here, and this particular object is little more than a private string reference that should never be used by other modules. I just built it this way because it made more semantic sense than `EC2_INSTACE_ID_ENV_KEY`

source/infrastructure/pipeline/e2e-tests/basic-ec2-start-stop.test.ts
```diff
@@ -0,0 +1,65 @@
...
+const instanceId = resourceParams.ec2InstanceId;
+
+test('instanceId exists', ()=> {
+  expect(instanceId).not.toBeNull()
+})
```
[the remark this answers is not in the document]
CrypticCabub: it's checking against undefined. I did not realize null and undefine were treated differently here.

source/infrastructure/pipeline/e2e-tests/basic-ec2-start-stop.test.resources.ts
```diff
@@ -0,0 +1,74 @@
...
+export const resourceParams = {
+  ec2InstanceId: process.env[envKeys.ec2InstanceId]!,
+  startStopTestScheduleName: 'ec2_basic_start_stop_test_schedule'
+}
```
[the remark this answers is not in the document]
CrypticCabub: typing. This should always be here when executing within the environment so asserting it as such makes using the value easier, but I added an extra test case that checks for `undefined` to explicitly catch the scenario where you accidently break the pipeline environment

source/infrastructure/pipeline/e2e-tests/basic-ec2-start-stop.test.resources.ts
```diff
@@ -0,0 +1,74 @@
...
+export class EC2StartStopTestResources implements TestResourceProvider {
+  createTestResources(scope: Construct) {
+
+    const vpc = new ec2.Vpc(scope, "basic-start-stop-vpc", {
+      natGateways: 0,
+      ipAddresses: ec2.IpAddresses.cidr("10.0.0.0/16")
+    })
```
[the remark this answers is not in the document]
CrypticCabub: I would agree with that, and it would be easy enough to achieve by creating some sort of test-resources-common.ts file

source/infrastructure/pipeline/e2e-tests/basic-ec2-start-stop.test.resources.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    const testInstance = new ec2.Instance(scope, "basic-start-stop-instance", {
+      instanceType: ec2.InstanceType.of(ec2.InstanceClass.T2, ec2.InstanceSize.MICRO),
+      machineImage: ec2.MachineImage.latestAmazonLinux(),
+      vpc: vpc,
+    })
```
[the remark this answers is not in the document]
CrypticCabub: even when doing so goes against the style of the rest of the object literal? I would agree when the entire literal can be made that way, but doing it for 1 variable among many feels to me like it would cost more mental energy than doing it this way.

source/infrastructure/pipeline/e2e-tests/basic-ec2-start-stop.test.ts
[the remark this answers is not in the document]
hearde: Are you saying you need to change it? Because it's checking against `null` here

source/infrastructure/pipeline/e2e-tests/basic-ec2-start-stop.test.resources.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    cdk.Tags.of(testInstance).add("Schedule", resourceParams.startStopTestScheduleName )
```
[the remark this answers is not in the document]
CrypticCabub: "Schedule" is the hard-coded default. To make it match the actual parameter I would need to add the schedule key as an output to the main control stack to be able to affect it.

This would affect the outputs on the production stack as well, but that's probably a worthy tradeoff?

source/infrastructure/pipeline/e2e-tests/basic-ec2-start-stop.test.ts
CrypticCabub: This would be a good point of discussion to have with the team.

source/infrastructure/pipeline/e2e-tests/basic-ec2-start-stop.test.resources.ts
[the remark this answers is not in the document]
hearde: I don't see a problem with adding it as an export, but I'm not sure that's necessary - it's also a parameter, right? If you create that stack, you should know what the value is.

source/infrastructure/pipeline/e2e-tests/utils/ec2-test-utils.ts
CrypticCabub: a unit test for the unit testing util? I suppose we could though I'm not sure it's necessary.

I added the little if catch there to intercept the undefined error that I had originally been having and replacing it with a more useful debugging error should there every be any problems in the future.

source/infrastructure/pipeline/e2e-tests/utils/ec2-test-utils.ts
CrypticCabub: the bang assert is there to control the typing. For unit-test code I typically want it to be as simple and straightforward as possible, and any errors that happen can just cause the test to fail and print to console. No need to waste the cognitive complexity on extra error handling

source/infrastructure/pipeline/lib/pipeline-stack.ts
CrypticCabub: only to avoid hardcoding strings like this in the middle of a bunch of other code

source/infrastructure/pipeline/e2e-tests/utils/ec2-test-utils.ts
hearde: Yeah, I think we should get into the habit of testing all of our code. Even helpers that are used by tests. One-offs and very simple things might not be necessary (like your wait helpers in another file), but I have a feeling this function will be reused a lot.

source/infrastructure/pipeline/e2e-tests/utils/schedule-test-utils.ts
CrypticCabub: How would I go about doing that? These are utilities for simplifying the interface with dynamodb so any tests on them would essentially be another set of full e2e tests.

Since they are purely testing utilities that are depended on by all the other e2e tests that will use dynamo I don't see much benefit to testing them separately as any failures with them would cause cascade failures in all the other e2e tests, and a few more failing tests would probably just add to the noise?

source/infrastructure/pipeline/e2e-tests/utils/ec2-test-utils.ts
hearde: You are getting some kind of error handling either way. With bang you are just saying you're happy with errors about accessing properties of undefined. It feels like an abuse of the type system to me. For cases like this, you may get behavior you're happy with, but for values you pass around that have types that aren't strictly correct, you might run into strange situations later when a `string` results in a weird exception because it's not a string.

source/infrastructure/pipeline/lib/pipeline-stack.ts
hearde: But if it's not reused for anything, what's wrong with it being hard-coded?

source/infrastructure/pipeline/lib/pipeline-stack.ts
CrypticCabub: oops

source/infrastructure/pipeline/e2e-tests/utils/schedule-test-utils.ts
hearde: I'm not saying to add more e2e tests. I'm saying to add unit tests that run as part of the unit test/build stage. You're assuming that a bug in one of your testing utilities would lead to failing tests. What if you introduce a bug that gives bad information to your tests in such a way where they always pass?

The pipeline is also purely for testing, and it should also have unit tests, and probably integration tests.

source/infrastructure/pipeline/lib/pipeline-stack.ts
CrypticCabub: nothing, this is still hard-coded ;).

Defining it this way is just an attempt to treat it as a 'property' of the testing pipeline rather than an implementation detail buried in the weeds of everything else.

hearde: I'm not saying that's a bad idea, but be careful doing things like that prematurely because I think that's the kind of thing that leads to proliferation of module-level variables you can see all over the application code. I think I would be a little more comfortable if they were instance variables of the pipeline stack class. I don't think them not being exported is as restrictive because I can always go in and `export` that variable, whereas I actually have to have pipeline stack instance to get at its instance variable.

source/infrastructure/pipeline/e2e-tests/utils/ec2-test-utils.ts
CrypticCabub: My assumption is that I am asserting the shape of a very specific return element from a larger api that can be used in other ways. I do not expect this value to ever be undefined except when the instanceID does not match up to a existing test instance. I could certainly be wrong about the behavior of the underlying interface, but I am basically attempting to take an interface build for batching and use it as a single-item lookup.

hearde (approved):

PR #412, opened 2023-03-23
Update README deployment instructions

CrypticCabub:
This is a followup PR to the end-to-end testing. The only relevant changes are to the readme as the rest are from  https://github.com/aws-solutions/aws-instance-scheduler/pull/410.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,26 +7,6 @@ Scheduler for Cross-Account and Cross-Region scheduling for EC2 and RDS instance
 To get started with Instance Scheduler, please review the solution documentation. 
 [Instance Scheduler on AWS](https://aws.amazon.com/solutions/implementations/instance-scheduler-on-aws/)
 
-## Building from GitHub
-***
-
-### Overview of the process
-
-Building from GitHub source will allow you to modify the solution. The process consists of downloading the source from 
-GitHub, creating buckets to be used for deployment, building the solution, and uploading the artifacts needed for deployment.
-
-#### You will need:
-
-* a Linux client with the AWS CLI v2 installed and python 3.7+, AWS CDK
-* source code downloaded from GitHub
-* two S3 buckets (minimum): 1 global and 1 for each region where you will deploy
-
-### Download from GitHub
-
-Clone the repository to a local directory on your linux client. Note: if you intend to 
-modify Instance Scheduler you may wish to create your own fork of the GitHub repo and work from that. 
-This allows you to check in any changes you make to your private copy of the solution.
-
 
 ## Repository Organization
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,44 +21,118 @@ This allows you to check in any changes you make to your private copy of the sol
     |- pipeline               - automated testing pipeline source files
 ```
 
-## Deploy the Solution From Github
+# Deploying the Solution
```
hearde, line 24: Usually a top-level heading is used for the title of the document

README.md
```diff
@@ -41,44 +21,118 @@ This allows you to check in any changes you make to your private copy of the sol
...
+* One global bucket that is accessed via the https end point. AWS CloudFormation templates are stored here. Ex. "mybucket"
 * One regional bucket for each region where you plan to deploy using the name of the global bucket as the root, and suffixed with the region name. Ex. "mybucket-us-east-1"
 * Your buckets should be encrypted and disallow public access
 
-**Build the solution**
+#### You will need:
+
+* a Linux client with the AWS CLI v2, python 3.7+, NPM, and Docker
+* a clone (or fork) of this repo
+* two S3 buckets (minimum): 1 global and 1 for each region where you will deploy
```
hearde, line 79: Do you really need a linux client? Don't we need a newer python version than 3.7? Specific node and npm versions? Docker meaning docker desktop or docker engine?

hearde (approved):

README.md
CrypticCabub: usually, though the layout of the readme currently has 2 sections that are arguably whole documents in themselves (deploying and testing) so that's why they are H1 level

README.md
CrypticCabub: the build script is a shell script that requires a linux-compatible terminal (as I know all too well >_>). the rest of it I'm not sure how to verify the actual versions which are needed but wanted to include some mention of them at least.

README.md
hearde: The title of the readme is the title of the solution. I would treat deploying and testing as second-level headings since they are both releated to the solution

README.md
hearde: If you're not sure, I wouldn't use specifics here. You don't have to specify the bare-minimum requirements, you can impose restrictions like "you should have the python3.9 because that's what the lambda runs on".

CrypticCabub: 3.7+ is what was in the file originally. Not sure where it came from but I think 3.9+ makes sense. alternatively I could just say "latest versions of" for both python and node

tmekari (changes requested):

README.md
```diff
@@ -41,50 +21,120 @@ This allows you to check in any changes you make to your private copy of the sol
     |- pipeline               - automated testing pipeline source files
 ```
 
-## Deploy the Solution From Github
+## Deploying the Solution
+***
+### One-Click Deploy From Amazon Web Services
+Refer to the [solution landing page](https://aws.amazon.com/solutions/implementations/instance-scheduler-on-aws/)
+to deploy Instance Scheduler on AWS using our pre-packaged deployment assets
```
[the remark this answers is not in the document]
CrypticCabub: updated

tmekari (approved):

PR #420, opened 2023-03-30
Remove unused files

hearde:
- remove unused files


source/app/Instance-scheduler.json
CrypticCabub, line 1: This file looks like it relates to defining CFN parameters, and several of the settings are suspiciously not the default. Is this potentially being used by nightswatch as the deployment parameters there??

schedulingActive is duplicated though (lines 3 and 7) so this can't possibly be valid, right?

CrypticCabub (approved):
Would like to run all these files by @gockle to confirm we aren't missing something, but I would agree with all of these appearing to be currently unused

PR #423, opened 2023-03-31
Changes for Aws Organizations, unit tests.

gockle:
*Issue #, if available:*

No Issues

*Description of changes:*
* Customer will be able to provide Aws Organization id instead of cross account roles
* [Design documentation](https://quip-amazon.com/CY4SAMk1UMaH/AWS-Instance-Scheduler-Scheduling-instances-across-multiple-accounts)
* Updated Unit tests for scheduler config files not comprehensive.
* Pending End to end test integration for cross account scheduling.

### DIFF
--- a/source/app/schedulers/instance_scheduler.py
+++ b/source/app/schedulers/instance_scheduler.py
@@ -107,8 +107,6 @@ class InstanceScheduler:
         self._context = None
         self._lambda_client = None
 
-
-
         self._usage_metrics = {"Started": {}, "Stopped": {}, "Resized": {}}
 
     @property
--- a/source/app/tests/test_scheduler_setup_handler.py
+++ b/source/app/tests/test_scheduler_setup_handler.py
@@ -20,4 +20,3 @@ def test_an_invalid_org_id_pattern():
     )
     response = handler.get_valid_org_id("111111111111,222222222222")
     assert response == None
-
--- a/source/app/tests/schedulers/test_instance_scheduler.py
+++ b/source/app/tests/schedulers/test_instance_scheduler.py
@@ -9,6 +9,7 @@ from schedulers import Ec2Service
 from util.named_tuple_builder import as_namedtuple
 from schedulers.instance_scheduler import InstanceScheduler
 
+
 @patch("boto3.Session")
 @patch("schedulers.Ec2Service")
 def test_scheduler_uses_regional_sts_endpoint(mock_ec2_service, mock_session):


source/infrastructure/instance-scheduler/lib/aws-instance-scheduler-remote-stack.ts
```diff
@@ -127,6 +153,82 @@ export class AwsInstanceSchedulerRemoteStack extends cdk.Stack {
...
+              {
+                EventBusName: mappings.findInMap("SchedulerEventBusName", "Name"),
+              }
+            ),
+          ],
+        }),
+      ],
+    });
+
+    // schedulerEventDeliveryRole.attachInlinePolicy(schedulerEventDeliveryPolicy)
```
CrypticCabub, line 190: commented code should be removed

source/infrastructure/instance-scheduler/lib/aws-instance-scheduler-remote-stack.ts
```diff
@@ -127,6 +153,82 @@ export class AwsInstanceSchedulerRemoteStack extends cdk.Stack {
...
+        "detail-type": ["Parameter Store Change"],
+        detail: {
+          name: ["/instance-scheduler/do-not-delete-manually"],
+          operation: ["Create", "Delete"],
+          type: ["String"],
+        },
+      },
+    });
+
+    // add condition to skip creation of these resources if deployed in hub account - Event Bus in the same account can not be used as target
```
CrypticCabub, line 220: I don't see this condition anywhere?

source/infrastructure/instance-scheduler/lib/event-bus-stack.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export interface SchedulerEventBusProps {
+  organizationId: string[];
+  namespace: string;
+  lambdaFunctionArn: string;
+  eventBusName: string;
+  isMemberOfOrganizationsCondition: cdk.CfnCondition;
+}
+
+export class SchedulerEventBusResources extends Construct {
```
CrypticCabub, line 16: should this be a capital C Construct or just a helper class like I did for core-scheduler.ts? @hearde your thoughts?

source/app/schedulers/instance_scheduler.py
```diff
@@ -180,21 +180,15 @@ def get_session_for_account(cross_account_role, aws_account):
...
-            account = role_elements[4]
+        # iterate through remote accounts
+        for account in self._configuration.remote_account_ids:
             if account in accounts_done:
-                self._logger.warning(WARN_DUPLICATE_ACCOUNT, account, role)
+                self._logger.warning(WARN_DUPLICATE_ACCOUNT, account)
                 continue
 
             # get a session for the role
+            # arn:aws:iam::597682407414:role/aws-ec2-spot-fleet-tagging-role
```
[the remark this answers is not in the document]
CrypticCabub: don't forget to remove the viperlight ignore at the same time

CrypticCabub (changes requested):

source/app/schedulers/instance_scheduler.py
```diff
@@ -180,21 +180,15 @@ def get_session_for_account(cross_account_role, aws_account):
...
+            role = f"arn:{self._configuration.aws_partition}:iam::{account}:role/{self._configuration.namespace}-{self._configuration.scheduler_role_name}"
```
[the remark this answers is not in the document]
CrypticCabub: is it? don't we also have to comply with this pattern in the remote stack when we create the ARN?

source/app/requesthandlers/eventbus_request_handler.py
```diff
@@ -0,0 +1,106 @@
...
+            detail = self._event.get("detail", None)
+            if detail is not None:
+                self._logger.debug(f"Details of the event {detail}")
+                dynamodb_table = DynamoDBUtils.get_dynamodb_table_resource_ref(
+                    self._config_table_name
+                )
+                config_key = {"name": "scheduler", "type": "config"}
+                if detail.get("operation") == EVENT_CREATE:
+                    self._logger.info(f"Add account id from the config")
+                    account = self._event.get("account")
```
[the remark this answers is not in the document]
CrypticCabub: That helps, but for things like this going forward I would really like to see an explicit object that describes this shape. Maybe it's not possible in the current architecture since we don't have typing setup at all?

source/app/requesthandlers/eventbus_request_handler.py
```diff
@@ -31,9 +29,11 @@ def __init__(self, event, context):
 
         # Setup logging
         self._is_trace_enabled = os.getenv("TRACE", False)
-        logging_stream_name = "-".join([LOG_STREAM_PREFIX])
+        logging_stream_name = "-".join(["eventbus_request_handler"])
         dt = datetime.utcnow()
-        logstream = LOG_STREAM.format(logging_stream_name, dt.year, dt.month, dt.day)
+        logstream = "{}-{:0>4d}{:0>2d}{:0>2d}".format(
+            logging_stream_name, dt.year, dt.month, dt.day
+        )
```
CrypticCabub: maybe I'm too function-happy, but I typically would lean toward including something like this as a separately named function entirely. This works fine as is, but I would be curious as to the team's opinion on code style best practices for stuff like this

source/app/requesthandlers/scheduler_setup_handler.py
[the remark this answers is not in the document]
CrypticCabub: as opposed to just storing environment variables on the lambda?

source/app/schedulers/instance_scheduler.py
```diff
@@ -119,6 +121,16 @@ def __init__(self, service, scheduler_configuration):
...
+    @property
+    def lambda_client(self):
+        """
+        Get the lambda client
+        :return: lambda client
+        """
+        if self._lambda_client is None:
+            self._lambda_client = get_client_with_standard_retry("lambda")
+        return self._lambda_client
+
```
CrypticCabub, line 133: we seem to follow this lazy-load pattern a lot. Not a change for right now, but if we're going to keep using this pattern I would be tempted to create an @lazy_load decorator to improve readability

source/app/schedulers/instance_scheduler.py
```diff
@@ -159,11 +171,53 @@ def get_session_for_account(cross_account_role, aws_account):
...
+                            Payload=payload,
+                        )
+                        self._logger.info(
+                            "Removing account {} from configuration".format(aws_account)
+                        )
+                        self._logger.debug(
+                            "Lambda response {} for removing account from configuration".format(
+                                response
+                            )
+                        )
```
CrypticCabub, line 207: there's a lot of code here in response to this AccessDenied error that I don't understand the purpose of. could this be extracted into a named function that explains what it does (or at the very least have a comment added)?

something like "reportFailure()" or "retryNameOfTask()"

source/app/tests/test_scheduler_setup_handler.py
```diff
@@ -2,14 +2,16 @@
...
+from unittest.mock import MagicMock
 
 
 def test_a_valid_org_id_pattern():
     handler = SchedulerSetupHandler(
         {"ResourceProperties": {"stack_version": "test"}}, {}
     )
     response = handler.is_valid_org_id("o-x1mhq1lvsr")
-    assert response != None
+    assert response != "o-x1mhq1lvsr"
```
CrypticCabub: this doesn't look right. the return type of is_valid_org_id is a string that matches the passed orgid?

source/app/tests/test_scheduler_setup_handler.py
```diff
@@ -18,3 +20,17 @@ def test_an_invalid_org_id_pattern():
...
+#         "configuration.config_admin.update_config", new_callable=mock.PropertyMock
+#     ) as mock_update_config:
+#         mock_update_config.return_value = None
+#         handler = SchedulerSetupHandler(
+#             {"ResourceProperties": {"stack_version": "test"}}, {}
+#         )
+#         handler.remote_account_ids = "o-x1mhq1lvsr"
+#         handler.use_aws_organizations = "True"
+#         response = handler._update_settings(prev_org_remote_account_ids=None)
+#         assert response == False
```
CrypticCabub, line 36: commented code

source/app/util/custom_resource.py
```diff
@@ -61,6 +61,11 @@ def request_id(self):
     def resource_properties(self):
         return self.event.get("ResourceProperties", {})
 
+    # Returns the previous resource properties of the custom resource, these are used to customize the updates
+    @property
+    def old_resource_properties(self):
+        return self.event.get("OldResourceProperties", {})
```
hearde, line 67: What custom resource is involved in this process? I thought it was an event from eventbridge

source/app/tests/test_scheduler_setup_handler.py
```diff
@@ -0,0 +1,36 @@
...
```
hearde, line 36: remove

source/app/requesthandlers/scheduler_setup_handler.py
```diff
@@ -222,22 +270,52 @@ def handle_request(self):
...
-    def _update_settings(self):
+    def is_valid_org_id(self, org_id):
+        """
+        Verifies if the ou_id param is a valid ou_id format. https://docs.aws.amazon.com/organizations/latest/APIReference/API_Organization.html
+        :return: boolean
+        """
+        try:
+            return re.fullmatch("^o-[a-z0-9]{10,32}$", org_id)
+        except Exception as error:
+            raise error
```
hearde, line 281: excepting then immediately raising is redundant

source/app/requesthandlers/scheduler_setup_handler.py
```diff
@@ -222,22 +270,52 @@ def handle_request(self):
...
+            except Exception as error:
+                self._logger.info(f"org id is not valid or empty {error}")
+                org_id = ""
+
+            if self.is_valid_org_id(org_id) and self.use_aws_organizations == "True":
+                self.organization_id = org_id
+                remote_account_ids = prev_org_remote_account_ids
+            elif self.is_valid_org_id(org_id) and self.use_aws_organizations == "False":
+                self.organization_id = org_id
+                remote_account_ids = {}
```
hearde, line 297: This section seems to check for the use aws organizations param being equal to "True" or "False" but the options in cdk are "Yes" and "No".

A way to avoid problems like this is to decouple the layer that understands the meaning of particular properties from the event object and translates them into Python-friendly objects for the rest of the code to use. Taking the whole event/context as a param means that every part of this class must understand exactly what the event looks like.

source/app/requesthandlers/scheduler_setup_handler.py
```diff
@@ -331,7 +409,31 @@ def _create_request(self):
...
+            and org_id == prev_org_id
+        ):
+            config = configuration.get_scheduler_configuration(self._logger)
+            prev_remote_account_id = config.remote_account_ids
+        else:
+            prev_remote_account_id = {}
+        return (
+            self._update_settings(prev_remote_account_id)
+            and self.set_lambda_logs_retention_period()
+        )
```
hearde, line 436: The control flow here is quite hard to follow. It would be nice to validate things earlier in the process so we don't have to be this paranoid later

source/app/schedulers/instance_scheduler.py
```diff
@@ -159,11 +171,53 @@ def get_session_for_account(cross_account_role, aws_account):
...
-                    ERR_ASSUMING_ROLE.format(cross_account_role, aws_account, str(ex))
+                    "Error Code {}".format(ex.response.get("Error", {}).get("Code"))
                 )
-                return None
+                if ex.response.get("Error", {}).get("Code") == "AccessDenied":
+                    try:
+                        self._logger.error(
+                            "Removing the account {} from scheduling configuration as assume role permission is missing for the iam role {}".format(
+                                aws_account, cross_account_role
+                            )
```
hearde, line 183: Is this the right thing to do? If the customer reinstalls the spoke stack will be the account be added back?

source/app/tests/test_instance_scheduler.py
```diff
@@ -1,13 +1,10 @@
 from unittest import mock
 import os
 from configuration.instance_schedule import InstanceSchedule
-from tests.botomock import mock_boto_env_vars
 from schedulers import Ec2Service
 from util.named_tuple_builder import as_namedtuple
 from schedulers.instance_scheduler import InstanceScheduler
 
-mock_boto_env_vars()
-
```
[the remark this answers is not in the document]
hearde: I don't see where it is... it's not an autouse fixture

source/app/tests/pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+addopts = -sv -p tests.plugins.env_vars
+log_cli = true
+log_level=INFO
```
hearde, line 4: are any of these actually necessary? Noisy test output obscures the results

source/app/tests/test_configuration_module.py
```diff
@@ -0,0 +1,311 @@
...
+                    )
+                    == "running"
+                )
+
+                invalid_datetime = datetime.datetime(2021, 6, 20, 17, 30, 59, 34258)
+                assert (
+                    period.get("period").get_desired_state(logger, invalid_datetime)
+                    == "stopped"
+                )
+
```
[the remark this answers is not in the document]
hearde: Having them together can provide value as well, to validate that it can make multiple scheduling decisions. but I agree that when the test fails it should be for one reason.

hearde (changes requested):
DIdn't review the actual logic too closely, just too large of a PR

source/app/requesthandlers/scheduler_setup_handler.py
hearde: @gockle this is the only thing I consider a blocking issue, I'll leave the rest up to your discretion

source/app/requesthandlers/scheduler_setup_handler.py
CrypticCabub: It may be redundant from a behavior perspective, but it does makes the ability of the code to throw an exception explicit rather than implicit, which I would personally consider to generally be a positive (whether or not that applies here specifically is up for some debate though).

source/infrastructure/instance-scheduler/lib/aws-instance-scheduler-remote-stack.ts
```diff
@@ -127,6 +153,82 @@ export class AwsInstanceSchedulerRemoteStack extends cdk.Stack {
       },
     ]);
 
+    // Event Rule to capture SSM Parameter Store creation by this stack
+    // SSM parameter to invoke event rule
+    const ssmParameterNamespace = new ssm.StringParameter(this, "SSMParameterNamespace", {
+      description: "This parameter is for Instance Scheduler solution to support accounts in AWS Organizations.",
+      stringValue: namespace.valueAsString,
+      parameterName: "/instance-scheduler/do-not-delete-manually",
+    });
```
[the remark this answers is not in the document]
hearde: I think I ran into this when attempting this same approach with WCO.

source/app/schedulers/instance_scheduler.py
CrypticCabub: if that's what it does, can a comment be added to explain that. My inability to understand this at a glance will likely apply to anybody coming back to read this code in the future.

source/app/tests/test_scheduler_setup_handler.py
CrypticCabub: if that's the behavior, then the function name should not be named "is_valid_org_id" the standard for "is_..." functions would be to return a true/false result.

hearde (approved):

CrypticCabub (approved):

PR #424, opened 2023-03-31
Custom Resource Fix

CrypticCabub:
Attempt to prevent custom resource from hanging and leaving CFN in a frozen state in the future (I just lost 2-3 hours to this freezing up my test environment >_>)

also includes a fix for sonarqube to correct internal pipeline

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -19,6 +19,7 @@ sonar.sources= \
 sonar.exclusions= \
     **/test/**, \
     **/tests/**, \
+    **/e2e-tests/**, \
     **/jest.config.ts, \
     bin/**, \
     source/app/certifi/**/*, \


sonar-project.properties
```diff
@@ -19,6 +19,7 @@ sonar.sources= \
 sonar.exclusions= \
     **/test/**, \
     **/tests/**, \
+    **/e2e-tests/**, \
```
hearde, line 22: Why are we excluding tests from sonarqube?

source/app/util/custom_resource.py
```diff
@@ -145,6 +145,9 @@ def handle_request(self):
         timer = threading.Timer(timeleft, self.fn_timeout)
```
hearde, line 145: Where are the unit tests for this file?

source/app/util/custom_resource.py
```diff
@@ -145,6 +145,9 @@ def handle_request(self):
...
         timer.start()
 
+        # default error to return if something causes the try to fail (this has been freezing for some reason)
+        self.response["Status"] = "FAILED"
+        self.response["Reason"] = "Unknown Error"
```
hearde, line 150: What problem does this solve specifically, and how?

A comment like this is worrying. We shouldn't be making changes to try random things. If we don't know how it works, we need to find out how and make it work the right way. I understand that is breaking the pattern in this solution, but that needs to change.

source/app/util/custom_resource.py
```diff
@@ -169,19 +172,20 @@ def handle_request(self):
         except Exception as ex:
             self.response["Status"] = "FAILED"
             self.response["Reason"] = str(ex)
-
-        timer.cancel()
-        return self._send_response()
+        finally:
+            timer.cancel()
+            return self._send_response()
```
hearde, line 177: I'm finding it hard to reason about the flow of this function. What I have done for custom resource providers is something like:

```py
try:
    # do work, raise Exception on error
    # send response with successful status
except:
    # send response with failed status
```

source/app/util/custom_resource.py
hearde, line 150: Bug: on completing a successful request, your response object contains the field `Reason` with the value `Unknown Error`.

hearde (changes requested):

sonar-project.properties
CrypticCabub: because the test code itself should not be included in the coverage metrics.

hearde: why not?

source/app/util/custom_resource.py
CrypticCabub: yeah good catch. is this change even worth pushing through at all right now? My goal was to try to add more ways for it to fail rather than failing silently (I was not able to find an error anywhere in the logs). I don't have enough information to know what caused the failure or how to make it repeatable/testable. So I see 3 options:
1 - do nothing, deal with it later
2 - refactor this custom resource 'properly' and add tests around the new design
3 - patch in some extra error reporting but otherwise leave the behavior alone (this is what I aimed for, but may have failed to achieve)

sonar-project.properties
CrypticCabub: It's standard practice as described in the sonarqube documentation [here](https://docs.sonarqube.org/9.8/project-administration/narrowing-the-focus/). Tests should be excluded from source scope and included under test scope instead

source/app/util/custom_resource.py
CrypticCabub: non existent, this is still legacy code. I'm not exactly confident that the timer works at all

PR #427, opened 2023-04-03
exclude e2e tests from sonar source directories

CrypticCabub:
[sonarqube's documentation](https://docs.sonarqube.org/9.8/project-administration/narrowing-the-focus/) specifies that tests should be excluded from the source directories and instead included as test directories explicitly. The e2e tests are already specified as a test directory, but were also being included as a source directory causing sonarqube to error when the files were being indexed twice. This change should bring us back into compliance with sonarqube standard practice as well as fix the error in the internal pipeline

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -19,6 +19,7 @@ sonar.sources= \
 sonar.exclusions= \
     **/test/**, \
     **/tests/**, \
+    **/e2e-tests/**, \
     **/jest.config.ts, \
     bin/**, \
     source/app/certifi/**/*, \


hearde (approved):
I'm not sure I agree with sonarqube's recommendation in principle. I can see justification for tracking test code coverage separately, and the top-level tests themselves don't need tests (for the most part...), but any helpers should probably have tests. But that's outside the scope of this review.

CrypticCabub:
> I'm not sure I agree with sonarqube's recommendation in principle. I can see justification for tracking test code coverage separately, and the top-level tests themselves don't need tests (for the most part...), but any helpers should probably have tests. But that's outside the scope of this review.

It literally errors out if you don't do it that way. Sonarqube isn't ignoring the tests entirely, I am just excluding it from the sources set. This directory is still imported into sonarqube under sonar.tests and I remember reading somewhere that sonar uses this to apply a different set of rules more appropriate for test code to that scope.

PR #428, opened 2023-04-03
Use regional STS endpoint to assume cross-account roles

CrypticCabub:
Acquire STS token from the regional endpoint of the region the lambda is currently executing in rather than the global endpoint. This causes STS to acquire a V2 token that is compatible with all regions instead of just standard regions.

manual test setup:
Hub stack deployed in US-East-1
Spoke stack deploy in Us-East 1

EC2 deployed Instances in spoke stack:
us-east-1 (same region): schedulable 
us-west-2 (cross-region, standard): schedulable
me-south-1 (cross-region, opt-in): schedulable
eu-central-2 (cross-region, boto filter error): not schedulable

this last one (eu-central-2) will be addressed in a separate PR covering the boto3 describe_regions filtration story


CrypticCabub:
Note: This change currently removes the boto3 automatic retries from the STS client used by the scheduler. the current boto3 retry implementation doesn't provide a way for me to add extra parameters to the function (namely the sts endpoint url). I was not confident enough in my understanding of the implementation of that utility to refactor it to support that, so have gone with the current approach for now.

source/app/schedulers/instance_scheduler.py
```diff
@@ -137,7 +137,16 @@ def _regions(self):
...
-            self._sts_client = get_client_with_standard_retry("sts")
+            session = boto3.Session()
+            sts_regional_endpoint = str.format(
+                "https://sts.{}.amazonaws.com", session.region_name
+            )
+            self._sts_client = session.client(
+                "sts",
+                region_name=session.region_name,
+                endpoint_url=sts_regional_endpoint,
+            )
```
hearde, line 148: If you're not going to use the helper to get a client, you need to be sure to give this the same behavior. Notably, you have not set the user agent extra string, nor have you configured retries.

source/app/schedulers/instance_scheduler.py
```diff
@@ -137,7 +137,16 @@ def _regions(self):
     @property
     def _sts(self):
         if self._sts_client is None:
-            self._sts_client = get_client_with_standard_retry("sts")
+            session = boto3.Session()
+            sts_regional_endpoint = str.format(
+                "https://sts.{}.amazonaws.com", session.region_name
+            )
```
hearde, line 143: Any pros/cons to getting the region name from the session rather than the lambda environment variable?

hearde (changes requested):

source/app/schedulers/instance_scheduler.py
CrypticCabub: which variable? I've not seen it documented anywhere and so don't trust it to be accurate at all. relying on the executing session seems safer from that perspective

source/app/schedulers/instance_scheduler.py
CrypticCabub: just said the same thing myself XD

if we need to maintain retries for this, I'll probably need a little help understanding the helper function. It's a little unclear exactly how the retries are added. (appears to me like it's set somewhere higher up the call chain)

hearde: No need to refactor the helper if that's difficult. Just specify retries and user agent extra here too.

source/app/schedulers/instance_scheduler.py
CrypticCabub: That's probably where boto is getting those values from anyway. One "pro" of this is that it's slightly easier to mock it directly in the test so I'll leave it as is.

source/app/schedulers/instance_scheduler.py
```diff
@@ -137,7 +137,16 @@ def _regions(self):
     @property
     def _sts(self):
         if self._sts_client is None:
-            self._sts_client = get_client_with_standard_retry("sts")
+            session = boto3.Session()
```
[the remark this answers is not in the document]
CrypticCabub: That's the region we want. this is not attempting to assume the region of the target, but rather the region where the lambda is being executed

source/app/schedulers/instance_scheduler.py
CrypticCabub: Took another look at it and refactored part of the helper such that I could extract the default retries configuration for re-use here

hearde (approved):

PR #431, opened 2023-04-11
Feature/dir structure

CrypticCabub:
Restructure the Lambda package to bundle instance_schedulers as a self-contained python module:

Update lambda environment to build with Poetry rather than Setup Tools

Fixes/improvements:

- lambda /build directory no longer grows recursively with every test execution due to being located within the main package
- lambda bundling now properly handled automagically by lambda construct (no editable install)
- dependencies now held constant by poetry.lock

all tests are passing


source/app/tests/test_main.py
```diff
@@ -0,0 +1,23 @@
...
+from instance_scheduler.main import lambda_handler
+
+
+class MockContext:
+    def __init__(self):
+        self.log_group_name = "My log group"
+
+
+@patch("instance_scheduler.main.SchedulerRequestHandler")
+def test_scheduler_request_handler_called(mock_handler):
```
hearde, line 14: Thanks for adding this back.

hearde (approved):
Thanks for fixing this up.

PR #433, opened 2023-04-11
Feature/scheduler cli build

CrypticCabub:
Scheduler CLI now builds with Poetry and integrates with the rest of the solution's testing framework.

changes:

- CLI now builds with poetry
- CLI integrates with Pytest/Tox/Sonarqube
- couple tests for basic operation
- handle case when user does not provide any input parameters (used to error, now prints command help)

closes: https://github.com/aws-solutions/aws-instance-scheduler/issues/323 https://github.com/aws-solutions/aws-instance-scheduler/pull/324 

Have run build-s3-dist locally and confirmed that the cli asset it produces installs correctly using `pip install .`


source/cli/instance-scheduler-cli-runner.py
```diff
@@ -15,7 +15,7 @@
 
 """Convenience wrapper for running instance-scheduler-cli directly."""
 
-from scheduler_cli.scheduler_cli import main
+from instance_scheduler_cli.scheduler_cli import main
 
 if __name__ == "__main__":
     main()
```
hearde, line 21: outside the scope of this review: surely this can be eliminated by adding something to `init`...

source/cli/tests/test_version.py
```diff
@@ -0,0 +1,5 @@
+from instance_scheduler_cli import __version__
+
+
+def test_version_correctly_picked_up_from_toml():
+    assert __version__ == "1.5.0"
```
hearde, line 5: Will this break when the version changes?

source/cli/pyproject.toml
```diff
@@ -0,0 +1,25 @@
...
+description = "AWS Instance Scheduler CLI"
+authors = ["AWS"]
+packages = [{ include="instance_scheduler_cli" }]
+
+[tool.poetry.scripts]
+scheduler-cli = 'instance_scheduler_cli.scheduler_cli:main'
+
+[tool.poetry.dependencies]
+python = "^3.9"
+#argparse = "^1.4.0" -- supposedly in the standard package now https://pypi.org/project/argparse/
```
hearde, line 17: Remove

source/cli/tests/test_version.py
CrypticCabub: yes, it will. we can remove this test entirely when that comes, but I figured a test breaking in that scenario was the lesser evil compared to not being certain that my changes to the "--version" command worked correctly.

source/cli/instance-scheduler-cli-runner.py
CrypticCabub: probably. I don't currently know enough of the nuance of python CLIs to be able to reason about why this should or should not exist

hearde (approved):

PR #447, opened 2023-04-14
CHORE: set sonarqube python version to 3.9

CrypticCabub:

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -58,6 +58,7 @@ sonar.python.coverage.reportPaths= \
      deployment/coverage-reports/cli-coverage.xml, \
      deployment/coverage-reports/lambda-coverage.xml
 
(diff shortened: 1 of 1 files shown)


sonar-project.properties
```diff
@@ -58,6 +58,7 @@ sonar.python.coverage.reportPaths= \
      deployment/coverage-reports/cli-coverage.xml, \
      deployment/coverage-reports/lambda-coverage.xml
 
+sonar.python.version=3.9
```
hearde, line 61: why not include 3.10 and 3.11, like tox.ini?

hearde (approved):

sonar-project.properties
CrypticCabub: because the lambda only runs on 3.9. 3.10 and 3.11 testing is included only on the github action as a forward-looking due-dilligence

PR #448, opened 2023-04-19
RDS e2e test

CrypticCabub:
**Add an e2e test for basic RDS start-stop**

Other changes:
- Include jest-extended library:
- test VPC now created by central helper function
- Schedule helper manages own client rather than being passed one with every call.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,7 +101,6 @@ ENV/
 
 #cdk 
 *cdk.out*
-*.d.ts
 *.js
 
 # config
--- a/source/infrastructure/pipeline/e2e-tests/utils/schedule-test-utils.ts
+++ b/source/infrastructure/pipeline/e2e-tests/utils/schedule-test-utils.ts
@@ -17,14 +17,19 @@ export interface Schedule {
   description: string;
   periods: Period[];
 }
-export async function createSchedule(client: dynamodb.DynamoDBClient, schedule: Schedule) {
-  await client.send(
-    new dynamodb.BatchWriteItemCommand({
-      RequestItems: {
-        [configTableName]: scheduleToBatchedPutRequests(schedule),
-      },
-    })
-  );
+export async function createSchedule(schedule: Schedule) {
+  const dynamoClient = new dynamodb.DynamoDBClient({});
+  try {
+    await dynamoClient.send(
+      new dynamodb.BatchWriteItemCommand({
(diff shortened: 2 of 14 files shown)


source/infrastructure/pipeline/e2e-tests/utils/schedule-test-utils.ts
```diff
@@ -17,14 +17,19 @@ export interface Schedule {
...
+    await dynamoClient.send(
+      new dynamodb.BatchWriteItemCommand({
+        RequestItems: {
+          [configTableName]: scheduleToBatchedPutRequests(schedule),
+        },
+      })
+    );
+  } finally {
+    dynamoClient.destroy();
+  }
```
CrypticCabub, line 32: removed need to pass dynamoclient with each schedule request, instead opting to create and destroy clients with each call.

This should help ensure that the clients get cleaned up properly as the number of tests grow (each test will create a new schedule, likely using a different client).

source/infrastructure/pipeline/e2e-tests/utils/vpc-utils.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const testVpcs = new Map<Construct, ec2.Vpc>();
+
+/**
+ * create a default testVPC for a given scope
+ * <p>
+ *   if a vpc has already been defined for this scope, it will be returned instead of creating a new one
+ * </p>
+ */
+export function defaultTestVPC(scope: Construct): ec2.Vpc {
```
CrypticCabub, line 16: VPCs are needed for all EC2 and RDS instances. This central utility will allow VPCs to be re-used while still allowing multiple VPCs to be created if in different scopes

.gitignore
```diff
@@ -101,7 +101,6 @@ ENV/
 
 #cdk 
 *cdk.out*
-*.d.ts
```
hearde, line 104: you can also leave the ignore as is and use an exclude to only not ignore this file: 
```
*.d.ts
!source/infrastructure/global.d.ts
```

hearde (approved):

.gitignore
CrypticCabub: valid, but I saw no reason to do this as the cdk output is no longer bundled under /source so this is, in theory, an incorrect exclusion under the "cdk" header

PR #497, opened 2023-11-16
fix: support china region

wchaws:
_Issue #, if available:_

_Description of changes:_ Support China region deployment. Since AppRegistry is not available in China region yet. Just disable AppRegistry creation if current region is China region. I've already tested in China region.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the
terms of your choice.


hearde:
Thank you for this contribution. We will integrate this change into an upcoming release.

PR #575, opened 2024-09-30
Release v3.0.5

CrypticCabub:
### Fixed
- Fixed bug in Nth weekday logic that would sometimes cause Nth weekday to be interpreted as 1 week too early

### Updated
- added rds:CreateDBSnapshot and rds:AddTagsToResource snapshot to scheduling roles to support recent changes to
RDS IAM requirements.

### Security
- Upgrade pyca/cryptography to mitigate GHSA-h4gh-qq45-vh27


abewub (approved):
approve v3.0.5

morjoan (approved):

